Issue #7, opened 2023-05-08
Show Reason syntax


jchavarri:
All snippets should be shown on either OCaml syntax or Reason syntax. We can write both manually if needed (it's not too much work), but some kind of switch should be added in the frontend.

For BuckleScript original site,  @rusty-key implemented the switch originally back in 2018: https://github.com/rescript-lang/bucklescript.github.io/pull/50

jchavarri:
When we tackle this task, we should also adapt these parts of docs to make sense when Reason syntax is selected:

- `bs` attribute as syntax sugar `f(. a, b)`
- Pipe operator (`|.` vs `->`)
- Some syntax like extensions differ (e.g. two percentage signs in `ml` are one in `re`)

As we mention these operators in paragraphs, maybe there's a way to use raw `<p>` elements with classnames, so that when switching syntaxes we show/hide these syntax-specific paragraphs.

PR #12, opened 2023-05-12
learn: communicate with JS

jchavarri:
Still some things missing, but I think the parts that are already there could be reviewed already.

### TODO

- [ ] Generators / `bs.deriving`
- [ ] `bs.scope`
- [ ] `bs.inline`
- [ ] `bs.optional`
- [ ] Document `unboxed` attribute for Melange
- [ ] Export to JS (`default` export)


docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+Here is the list of all extensions supported by Melange:
+
+- `bs.debugger`: insert `debugger` statements
+- `bs.external`: read global values
+- `bs.obj`: create JavaScript object literals
+- `bs.raw`: write raw JavaScript code
+- `bs.re`: insert regular expressions
+
+### Intro to `external`
```
anmonteiro, line 1052: should we call this section "foreign function interface" or "modeling foreign JavaScript code"?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+This is a final escape hatch which does nothing but convert from the type `foo`
+to `bar`. In the following sections, if you ever fail to write an `external`,
+you can fall back to using this one. But try not to.
+
+Let’s now see all the ways to use JavaScript from Melange.
+
+### Generate raw JavaScript
+
+It is possible to directly write JavaScript code from a Melange file. This is
+unsafe, but it can be useful for prototyping or as a escape hatch.
```
anmonteiro, line 1104:
```suggestion
unsafe, but it can be useful for prototyping or as an escape hatch.
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+  }
+|}]
+
+let () = Js.log (add 1 2)
+```
+
+The `{||}` strings are called ["quoted
+strings"](https://ocaml.org/manual/lex.html#quoted-string-id). They are similar
+to JavaScript’s template literals, in the sense that they are multi-line, and
+there is no need to escape characters inside the string.
```
anmonteiro, line 1123: but they don't support interpolation.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+is useful as we can attach the type signature already in the same line, to make
+our code safer. For example:
+
+```ocaml
+[%%bs.raw "var a = 1"]
+
+let f : unit -> int = [%bs.raw "function() {return 1}"]
+```
+
+Using two percentages `%%` (`[%%bs.raw <string>]`) is reserved for top-level
```
anmonteiro, line 1136:
```suggestion
Using two percentage signs (`[%%bs.raw <string>]`) is reserved for top-level
```

anmonteiro: not just top-level definitions, but rather structure items, e.g. 

```ocaml

module Foo = struct
  [%%bs.raw "not really toplevel"]
end
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+definitions, for example:
+
+```ocaml
+[%%bs.raw "var a = 1"]
+```
+
+### Debugger
+
+Melange allows to inject a `[%bs.debugger]` expression:
```
anmonteiro, line 1145:
```suggestion
Melange allows to inject a `debugger;` expression:
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+let () = match [%bs.external __filename] with
+| Some f -> Js.log f
+| None -> Js.log "non-node environment"
+```
+
+### Bind to JavaScript objects
+
+JavaScript objects are used in a variety of use cases:
+
+- As a fixed shape "record" or "struct" in other languages (like OCaml and C).
```
anmonteiro, line 1190: struct is a bit misleading here because of OCaml's `module X = struct .... end`. should we omit it?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+- As a map or dictionary.
+- As a class.
+- As a module to import/export.
+
+Melange cleanly separates the binding methods for JavaScript objects based on
```
anmonteiro, line 1195: cleanly seems unnecessary (compared to what?)
```suggestion
Melange separates the binding methods for JavaScript objects based on
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+```
+
+##### Using `Js.t` objects
+
+Alternatively to records, Melange offers another type that can be used to
+produce JavaScript objects. This type is `'a Js.t`, where `'a` is an [OCaml
+object](https://v2.ocaml.org/manual/objectexamples.html).
+
+The advantage of objects versus records is that no type declaration is needed in
+advanced, which can be helpful for prototyping or quickly generating JavaScript
```
anmonteiro, line 1286:
```suggestion
advance, which can be helpful for prototyping or quickly generating JavaScript
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+```js
+var john = {
+  name: "john",
+  age: 99
+};
+
+var t = john.name;
+```
+
+Note that type inference with objects is more lenient than with records. For
```
anmonteiro, line 1309: it's not "more lenient", it's structural (with optional row polymorphism) vs nominal typing.

I understand it might be harder to explain, but "more lenient" is a bit weird to say.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+include a field `name` that is of type string, e.g.:
+
+```ocaml
+let name_extended obj = obj##name ^ " wayne"
+
+let one = name_extended [%bs.obj { name = "john"; age = 99 }]
+let two = name_extended [%bs.obj { name = "jane"; address = "1 infinite loop" }]
+```
+
+This is possible thanks to [objects
```
anmonteiro, line 1320: this is possible thanks to structural typing rather than polymorphism.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+polymorphic variants. The book Real World OCaml has [a good
+introduction](https://dev.realworldocaml.org/objects.html) to object
+polymorphism.
+
+##### Bind to object properties
+
+If you need to bind only to the property of a JavaScript object, you can use
+`bs.get` and `bs.set`:
+
+```ocaml
```
anmonteiro, line 1332: we should also include an example of what this codes generates, so that people can distinguish between `foo.bar` and `foo["bar"]`

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+```
+
+#### Objects with dynamic shape (dictionary-like)
+
+If you are binding to a JavaScript object that:
+
+- might or might not add or remove keys
+- contains only values that are of the same type
+
+Then it’s not really an object, it’s a dictionary or map. Use
```
anmonteiro, line 1368: let's say "then you can model it in OCaml using `Js.Dict`."

> "it's not really an object"

but, it is :P

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+The generated JavaScript function is the same, but now the usage in Melange is
+much clearer.
+
+**Note**: in this particular case, a final param of type unit, `()` must be
+added after `border`, since `border` is an optional argument at the last
+position. Not having the last param `unit` would lead to a warning, which is
+explained in detail [in the OCaml
+documentation](https://ocaml.org/docs/labels#warning-this-optional-argument-cannot-be-erased).
+
+Note that you can freely reorder the labels on the Melange side, they will
```
anmonteiro, line 1491: on the application side. but the definition side must match the order of arguments in JS. I think it's important to mention that.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+[in the OCaml Cornell
+textbook](https://cs3110.github.io/textbook/chapters/modules/encapsulation.html)
+provides a great overview to understand what abstract types are and when they
+are useful.
+
+##### Chaining
+
+It is common to find this kind of API in JavaScript: `foo().bar().baz()`. This
+kind of API can be designed with Melange externals and the `bs.send` attribute,
+in combination with the [the pipe operator](todo-fix-me.md).
```
anmonteiro, line 1544: feels like an example is missing!

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+  [@@bs.send]
+
+let el = get_by_id document "my-id"
+```
+
+```js
+var el = document.getElementById("my-id");
+```
+
+When using `bs.send`, the first argument will be the object that holds the
```
anmonteiro, line 1531: should we also document `bs.send.pipe`, which inserts the argument at the end and is more suitable to be used with `|>`?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+It is common to find this kind of API in JavaScript: `foo().bar().baz()`. This
+kind of API can be designed with Melange externals and the `bs.send` attribute,
+in combination with the [the pipe operator](todo-fix-me.md).
+
+#### Variadic function arguments
+
+Sometimes JavaScript functions take an arbitrary amount of arguments. For this
+cases, Melange provides the `bs.variadic` attribute, which can be attached to
+the `external` declaration. However, there is one caveat: all the variadic
+arguments need to belong to the same type.
```
anmonteiro, line 1551: should we document the solution for this?

```ocaml
type hide = Hide : 'a -> hide [@@unboxed]

external join : hide array -> string = "join" [@@bs.module "path"] [@@bs.splice]

let v = join [| Hide "a"; Hide 2 |]

```

which generates:

```js
// Generated by Melange
'use strict';

var Path = require("path");

var v = Path.join("a", 2);

exports.v = v;
/* v Not a pure module */
```

The same trick can be used for `Js.Dict`s

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Generates:
+
+```js
+var Path = require("path");
+var v = Path.join("a", "b");
+```
+
+#### Bing to a polymorphic function
+
+Some JavaScript libraries will define functions that where the arguments can
```
anmonteiro, line 1568:
```suggestion
Some JavaScript libraries will define functions where the arguments can
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+vary on both type and shape.There are two approaches to bind to those, depending
```
anmonteiro, line 1569:
```suggestion
vary on both type and shape. There are two approaches to bind to those, depending
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+  throw new Error(`Expected string or number, got '${padding}'.`);
+}
+```
+
+Here, `padding` can be modeled as a polymorphic variant.
+
+```ocaml
+external padLeft:
+  string
+  -> ([ `Str of string
```
anmonteiro, line 1611: requirements:

- needs to be a polyvariant
- needs to contain payload

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Node’s `fs.readFileSync` second argument. It can only take a few given string
+values: `"ascii"`, `"utf8"`, etc.
+
+One could still bind it just as a string, but this would not prevent from using
+unsupported values. Let’s see how we can use polymorphic variants and the
+`bs.string` attribute to avoid runtime errors:
+
+```ocaml
+external read_file_sync :
+  name:string -> ([ `utf8 | `ascii ][@bs.string]) -> string = "readFileSync"
```
anmonteiro, line 1645: I think this is not needed anymore since polyvariants compile to strings.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+external test_int_type :
+  ([ `on_closed | `on_open [@bs.as 20] | `in_bin ][@bs.int]) -> int
+  = "testIntType"
+  [@@bs.val]
+
+let value = test_int_type `on_open
+```
+
+In this example, `on_closed` will be encoded as 0, `on_open` will be 20 due to
+the attribute `bs.as` and `in_bin` will be 21.
```
anmonteiro, line 1676: we need to explain why: because it starts counting up from the previous value.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+```ocaml
+let add x y = x + y
+```
+
+Its type will be `int -> int -> int`. This means that one can partially apply
+`add` by calling `add 1`, which will return another function expecting the
+second argument of the addition. This kind of functions are called "curried"
+functions, more information about currying in OCaml can be found in [this
+chapter](https://cs3110.github.io/textbook/chapters/hop/currying.html) of the
```
anmonteiro, line 1757: I wonder if we should compile all these external resources presented across the website and also expose them in a dedicated section.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+"OCaml Programming: Correct + Efficient + Beautiful" book.
+
+Currying is problematic when binding to JavaScript, let’s say we have a function
```
anmonteiro, line 1760: I don't like the word "problematic". We should probably just say "incompatible calling conventions", no?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Now if we ever used our external function `map` with our `add` function by
+calling `map arr1 arr2 add` it would not work as expected. JavaScript function
+application does not work the same as in OCaml, so the function call in the
+`map` implementation, `f(a[i],b[i])`, would be applied over the outer JavaScript
+function `add`, which only takes one argument `x`, and `b[i]` would be just
+discarded. The value returned from the operation would not be the addition of
+the two numbers, but rather the inner anonymous callback.
+
+To solve this mismatch between OCaml and JavaScript functions and their
+application, Melange provides a special attribute `@bs` that can be used with
```
anmonteiro, line 1811: should we also mention the ReasonML syntax sugar? `f(. a, b)`.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+In the example above:
+
+```ocaml
+external map : 'a array -> 'b array -> (('a -> 'b -> 'c)[@bs]) -> 'c array
+  = "map"
+  [@@bs.val]
+```
+
+Here `('a → 'b → 'c [@bs])` will be interpreted as having arity 2, in general,
```
anmonteiro, line 1822: what's this `→` character? can we use `->` instead?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+conversion functions like `Js.Nullable.toOption` and such.
+
+```ocaml
+type element
+type document
+external get_by_id : document -> string -> element option = "getElementById"
+  [@@bs.send] [@@bs.return nullable]
+
+let test document =
+  let elem = document |. get_by_id "header" in
```
anmonteiro, line 1932: have we introduced `|.` before? This will be unfamiliar to people if we haven't.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+nullable]` above. Currently 4 directives are supported: `null_to_opt`,
+`undefined_to_opt`, `nullable` and `identity`.
+
+`nullable` is encouraged, as it will convert from `null` and `undefined` to
+`option` type.
+
+<!-- When the return type is unit: the compiler will append its return value with an OCaml unit literal to make sure it does return unit. Its main purpose is to make the user consume FFI in idiomatic OCaml code, the cost is very very small and the compiler will do smart optimizations to remove it when the returned value is not used (mostly likely). -->
+
+`identity` will make sure that compiler will do nothing about the returned
+value. It is rarely used, but introduced here for debugging purpose.
```
anmonteiro, line 1961:
```suggestion
value. It is rarely used, but introduced here for debugging purposes.
```

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
 
 ## Communicate with JavaScript
 
-_TODO: Sections will be something like:_
+Melange offers an extensive range of techniques to use JavaScript code in
+Melange projects and to make Melange code accessible to JavaScript programs. To
+learn about these techniques, we will go through the types in Melange and how
+they map to JavaScript runtime code, then we will see what are the OCaml
```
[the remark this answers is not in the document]
jchavarri: I took this suggestion and adapted it to surrounding text in https://github.com/melange-re/documentation-site/pull/12/commits/77b246e1cb80d752e0f0c2b1f297a8484cb54847.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+construct. Finally, we will provide a variety of use cases with examples to show
+how to communicate to and from JavaScript.
+
+### Data types and runtime representation
+
+This is how each Melange type is converted into JavaScript values:
+
+Melange | JavaScript
+---------------------|---------------
+int | number
```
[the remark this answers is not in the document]
jchavarri: Do you think any further note would be useful here?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Js.Nullable.t | `null` / `undefined`
+Option.t `None` | `undefined`
+Option.t `Some( Some .. Some (None))` | internal representation
+Option.t `Some 2` | `2`
+record `{x: 1; y: 2}` | object `{x: 1, y: 2}`
+int64 | array of 2 elements `[high, low]` high is signed, low unsigned
+char | `'a'` -> `97` (ascii code)
+bytes | number array
+list `[]` | `0`
+list `[x, y]` | `[x, [y, 0]]`
```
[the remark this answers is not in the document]
jchavarri: Fixed in https://github.com/melange-re/documentation-site/pull/12/commits/c9fa4fe982809950f537bca9ffd590e1c0d78295 (also the line below).

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+```
+
+Polymorphic Variant:
+
+```ocaml
+let u = `Foo (* "Foo" *)
+let v = `Foo(2) (* { NAME: "Foo", VAL: "2" } *)
+```
+
+> **_NOTE:_** Relying on runtime representations by reading or writing them
```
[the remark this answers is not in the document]
jchavarri: > but sometimes it's necessary. We shouldn't be so stern here.

Would you remove the note altogether, or reword it? The intent of the note was to made clear that it is possible, but dangerous from a typing perspective (which I think it is, same as writing bindings).

> Perhaps we need to document which representations are stable and which ones aren't.

This is documented already:

https://github.com/melange-re/documentation-site/blob/c9fa4fe982809950f537bca9ffd590e1c0d78295/docs/learn.md?plain=1#L676

https://github.com/melange-re/documentation-site/blob/c9fa4fe982809950f537bca9ffd590e1c0d78295/docs/learn.md?plain=1#L757

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Let’s see now some of these types in detail.
+
+### Shared types
+
+The following are types that can be shared between Melange and JavaScript almost
+"as is". Specific caveats are mentioned on the sections where they apply.
+
+#### Strings
+
+Melange strings are an immutable sequence of bytes. This is problematic when
```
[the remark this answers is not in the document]
jchavarri: Reworded in https://github.com/melange-re/documentation-site/pull/12/commits/f41a0fa018ed50b2529a571e9ee466a6a8652447.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+define [quoted string
+literals](https://v2.ocaml.org/manual/lex.html#sss:stringliterals) using the
+`js` identifier, for example:
+
+```ocaml
+let () = Js.log {js|你好，
+世界|js}
+```
+
+For convenience, Melange exposes another special quoted string identifier: `j`.
```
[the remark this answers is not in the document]
jchavarri: It is documented later on:

https://github.com/melange-re/documentation-site/blob/c9fa4fe982809950f537bca9ffd590e1c0d78295/docs/learn.md?plain=1#L1624-L1625

I think it's better to keep `{json||json}` in context, as it is a pretty niche feature. Wdyt?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+#### Float
+
+Melange floats are JavaScript numbers, and vice-versa. The Melange standard
+library doesn’t come with a Float module. The JavaScript Float API can be found
+in the [`Js.Float`](todo-fix-me.md) module.
+
+#### Int
+
+**Ints are 32-bits**. They compile to JavaScript numbers but treating both as
```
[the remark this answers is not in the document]
jchavarri: I tried to adapt the text, but then a few lines below we mention:

> treating both as the same might lead to unexpected behavior, because of the differences in precision

as JS ints might get up to 2^53 - 1. So we would need to explain how JavaScript can adapt the integers depth dynamically based on the operations applied to them.

I fear it'd be too much detail to include these JS details for the more generic / entry level manual to Melange. Wdyt?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+the same might lead to unexpected behavior, because of the differences in
+precision. When working with large numbers, it is better to use floats. For
+example, the bindings in `Js.Date` use floats. The bindings to work with
+JavaScript integers are in the [`Js.Int`](todo-fix-me.md) module.
+
+#### Array
+
+Idiomatic OCaml arrays are supposed to be fix-sized. This constraint is relaxed
+on the Melange side. You can change its length using the usual [JavaScript Array
```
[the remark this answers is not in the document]
jchavarri: > is that actually true? 

Changing the length is possible. The linked issue only allowed to do so in a safe way, right?

I rewrote this section in https://github.com/melange-re/documentation-site/pull/12/commits/e6c23fc2ea7f4b223899f2ea648d13b2fb2e39eb, to include also homomorphism of Melange arrays as a difference against JavaScript arrays.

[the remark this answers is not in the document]
jchavarri: https://github.com/melange-re/documentation-site/pull/12/commits/e6c23fc2ea7f4b223899f2ea648d13b2fb2e39eb

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+library doesn’t come with a Float module. The JavaScript Float API can be found
+in the [`Js.Float`](todo-fix-me.md) module.
+
+#### Int
+
+**Ints are 32-bits**. They compile to JavaScript numbers but treating both as
+the same might lead to unexpected behavior, because of the differences in
+precision. When working with large numbers, it is better to use floats. For
+example, the bindings in `Js.Date` use floats. The bindings to work with
+JavaScript integers are in the [`Js.Int`](todo-fix-me.md) module.
```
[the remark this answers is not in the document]
jchavarri: Added in https://github.com/melange-re/documentation-site/pull/12/commits/e6c23fc2ea7f4b223899f2ea648d13b2fb2e39eb.

docs/learn.md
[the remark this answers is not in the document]
jchavarri: Actually, I changed my mind and ended up adding explanation about all this in https://github.com/melange-re/documentation-site/pull/12/commits/e6c23fc2ea7f4b223899f2ea648d13b2fb2e39eb. 🤷‍♂️

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+好，$(world)|j}`.
+
+To work with strings, Melange standard library provides some utilities in the
+[`Stdlib.String` module](todo-fix-me.md). The bindings to the native JavaScript
+functions to work with strings are in the [`Js.String` module](todo-fix-me.md).
+
+#### Float
+
+Melange floats are JavaScript numbers, and vice-versa. The Melange standard
+library doesn’t come with a Float module. The JavaScript Float API can be found
```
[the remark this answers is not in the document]
jchavarri: https://github.com/melange-re/documentation-site/pull/12/commits/e6c23fc2ea7f4b223899f2ea648d13b2fb2e39eb

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+#### Array
+
+Idiomatic OCaml arrays are supposed to be fix-sized. This constraint is relaxed
+on the Melange side. You can change its length using the usual [JavaScript Array
+API](todo-fix-me.md). Melange’s functions to work with arrays are in the
+[`Js.Array`](todo-fix-me.md) module.
+
+#### Tuple
+
+OCaml tuples are compiled to JavaScript arrays. This is convenient when writing
```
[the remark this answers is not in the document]
jchavarri: I added an example in https://github.com/melange-re/documentation-site/pull/12/commits/8980c2eccab3f91f0462a2be188964406127a725. Please let me know if it was what you had in mind.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+Fortunately, OCaml provides mechanisms to extend its language without breaking
+compatibility with the parser or the language. These mechanisms are composed by
+two parts:
+- First, some syntax additions to define parts of the code that will be extended
+  or replaced
+- Second, compile-time OCaml native programs called [PPX
+  rewriters](https://ocaml.org/docs/metaprogramming), that will read the syntax
+  additions defined above and proceed to extend or replace them
+
+The syntax additions come in two flavours, called [extensions
```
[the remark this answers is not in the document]
jchavarri: https://github.com/melange-re/documentation-site/pull/12/commits/5117a37a47e517cb6b0cb06268f42771367eebfa

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+```js
+var a = 1; var b = 2
+var add = a + b
+```
+
+The difference between one and two `%` characters is detailed in the [OCaml
+documentation](https://v2.ocaml.org/manual/extensionnodes.html).
+
+Attributes are "lighter" than extensions, and are used to annotate specific
```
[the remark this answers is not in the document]
jchavarri: And https://github.com/melange-re/documentation-site/pull/12/commits/8bd60df88f6c0fffc6470361bca39ef570e7aed2 😓

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+> **_NOTE:_** All these attributes and extensions are prefixed with `bs.` for
+> backwards compatibility. They will be updated to `mel.` in the future.
+
+**Attributes:**
+
+Used to annotate `external` definitions:
+
+- [`bs.get`](#bind-to-object-properties): read JavaScript object properties
+  statically by name
```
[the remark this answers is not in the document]
jchavarri: I am not sure I understand what the suggestion is, could you elaborate?

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+- [`bs.get_index`](#bind-to-object-properties): read a JavaScript object’s
+  properties dynamically by using the bracket notation `[]`
+- [`bs.module`](#using-functions-from-other-javascript-modules): bind to a value
+  from a JavaScript module
+- [`bs.new`](#javascript-classes): bind to a JavaScript constructor
+- [`bs.obj`](#using-jst-objects): create JavaScript object
+- [`bs.return`](#wrapping-returned-nullable-values): automate conversion from
+  nullable values to `Option.t` values
+- [`bs.send`](#calling-a-function-as-an-object-property): call a function that
```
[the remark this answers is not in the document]
jchavarri: Yes, much better. Added in https://github.com/melange-re/documentation-site/pull/12/commits/28907dc02a0e4cc37a2ddc6d2502db5c777c1dd8, plus tweaked the titles in `bs.send` section.

docs/learn.md
jchavarri: Agree. Renamed in https://github.com/melange-re/documentation-site/pull/12/commits/d3da08456c9e3cdddb02a84e1348633f764e342f.

docs/learn.md
jchavarri: But some of the quoted strings can interpolate :P https://reasonml.github.io/en/try?rrjsx=true&reason=DYUwLgBJC8EEQDMD2S4G4BQHSQK4VgG8ArAHwAkRhgkIASMU4gXyA

I am not sure how to tackle this suggestion, I think it'd be better to leave it as is, and not explain any more that these two similarities (multi-line and escaping), but if you have any ideas on how to explain the interpolation limitation while explaining why `{j||j}` does support it, happy to add them.

docs/learn.md
jchavarri: I did not include this particular example for simplicity. But removed the "top-level" reference, mentioned structure and signature definitions, and linked to OCaml manual in https://github.com/melange-re/documentation-site/pull/12/commits/787e2619e9ca835178c61918f01b9ad45ed97eac.

docs/learn.md
jchavarri: Yeah, removed in https://github.com/melange-re/documentation-site/pull/12/commits/3fb7df5d8288cee48739f79312e06acaf6a5e048.

(TIL that C# has [record structs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/record-structs)! #allthethings)

docs/learn.md
jchavarri: Agree, in general I have been trying to remove as many adjectives as possible, and stick to purely descriptive language.

Removed in https://github.com/melange-re/documentation-site/pull/12/commits/3fb7df5d8288cee48739f79312e06acaf6a5e048.

docs/learn.md
jchavarri: I tweaked this part of the docs in https://github.com/melange-re/documentation-site/pull/12/commits/65a43aac373be89fafe069f10dc2926b94342126.

docs/learn.md
jchavarri: I tweaked this section in https://github.com/melange-re/documentation-site/pull/12/commits/65a43aac373be89fafe069f10dc2926b94342126.

docs/learn.md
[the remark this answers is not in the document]
jchavarri: I got it now, added a reference to "dot notation" following [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors, and example in the corresponding section: https://github.com/melange-re/documentation-site/pull/12/commits/54cc220edc62adaa052f0295f2b3ee757d53eeea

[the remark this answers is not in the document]
jchavarri: Let me know if this is enough. Note this section idea was to be as brief as possible (mostly enumeration of the available attrs and extensions and 1-line sentence to define them).

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: probably not. good as is.

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: ok then nothing actionable?

docs/learn.md
jchavarri: Rewrote this section fully in https://github.com/melange-re/documentation-site/pull/12/commits/019bf1a6c0d9429bdb97737f375f39471183cc80, please lmk what you think.

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: fine with me.

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: > integers themselves have a [max depth of 53

this is wrong, after 2^32-1 they're floating point.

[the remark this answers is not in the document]
anmonteiro: the rest looks good though.

docs/learn.md
```diff
@@ -721,20 +723,1250 @@ template](https://github.com/melange-re/melange-opam-template).
...
+
+**Ints are 32-bits**. They compile to JavaScript numbers but treating both as
+the same might lead to unexpected behavior, because of the differences in
+precision. When working with large numbers, it is better to use floats. For
+example, the bindings in `Js.Date` use floats. The bindings to work with
+JavaScript integers are in the [`Js.Int`](todo-fix-me.md) module.
+
+#### Array
+
+Idiomatic OCaml arrays are supposed to be fix-sized. This constraint is relaxed
```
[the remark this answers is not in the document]
anmonteiro: there's still "fix-sized" instead of "fixed-size"

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: looks good.

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: yeah, nice.

docs/learn.md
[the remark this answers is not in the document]
jchavarri: I tried to tackle your original comment by rewriting the paragraph and removing parts of it in https://github.com/melange-re/documentation-site/pull/12/commits/b2e6ef741b09c1053470b0626d40b3358d6b49ba.

docs/learn.md
anmonteiro: I don't know if we should mention subtyping, since we don't give examples for it. This is about structural typing vs nominal typing.

docs/learn.md
anmonteiro: very nice!

docs/learn.md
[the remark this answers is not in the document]
anmonteiro: nice, that reads a lot better.

docs/learn.md
jchavarri: Fixed in https://github.com/melange-re/documentation-site/pull/12/commits/4d2c9ba756c3c28304e12ff8eacbc092a8ccd9ab.

docs/learn.md
jchavarri: Added in https://github.com/melange-re/documentation-site/pull/12/commits/7e661312d55c4a100e317f2ae7cf2baca071ca66

docs/learn.md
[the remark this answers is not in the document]
jchavarri: cool, resolving this conversation then (for the sake of tracking remaining threads)

docs/learn.md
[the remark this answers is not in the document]
jchavarri: resolving as agreed offline that the new changes make sense.

docs/learn.md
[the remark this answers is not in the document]
jchavarri: Gave another try in https://github.com/melange-re/documentation-site/pull/12/commits/1484685819554d08386a6ab357c482e7265bfee7, lmk what you think.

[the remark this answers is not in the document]
jchavarri: And https://github.com/melange-re/documentation-site/pull/12/commits/cc4423bfe041f07eddd8b9b0b7448c13b506296c.

docs/learn.md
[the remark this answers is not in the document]
jchavarri: Oops. Fixed in https://github.com/melange-re/documentation-site/pull/12/commits/630615e1bc3421d52107b40b7d8fdb0292488792.

docs/learn.md
jchavarri: removed refs to subtyping in https://github.com/melange-re/documentation-site/pull/12/commits/11740ebc41db0e86e170588969cc503af27b3736.

jchavarri:
I realized abstract types are crucial to writing bindings, but they were never documented explicitly, so added a small section for them in https://github.com/melange-re/documentation-site/pull/12/commits/d22f7fe639c6432bb8249ad7b0ba426d142bd6a8.

docs/learn.md
jchavarri: Added in https://github.com/melange-re/documentation-site/pull/12/commits/80c512d7e8fa2571e645db5ef693d1acbc394585, lmk what you think as it implied quite a bit of changes.

I referred to the blog post I wrote back in the day, maybe we could include other links that explain OCaml pipe `|>`, not sure.

docs/learn.md
jchavarri: Added solution for variadic in https://github.com/melange-re/documentation-site/pull/12/commits/b3bad3de602c738f17111b0e1689b0f3534ea247.

> The same trick can be used for `Js.Dict`s

The `unboxed` solution with existential types (is that what `hide` is?) for dictionaries would mean values can be added to the dictionary, but not be read from it, right?

docs/learn.md
jchavarri: Added in https://github.com/melange-re/documentation-site/pull/12/commits/11e3774e272a96fa359980d5fbaf4d76f7bd7a53, I also included the 2 other requisites.

docs/learn.md
jchavarri: I tried removing `bs.string` but it seems it makes using `bs.as` impossible, tracked in https://github.com/melange-re/melange/issues/578. So for now decided to leave it.

Added an example with `bs.string` and `bs.as` in https://github.com/melange-re/documentation-site/pull/12/commits/15499fb1f3e577830b4fc6e3a4e8c29d28d51b15.

docs/learn.md
jchavarri: Added in https://github.com/melange-re/documentation-site/pull/12/commits/15499fb1f3e577830b4fc6e3a4e8c29d28d51b15.

docs/learn.md
jchavarri: Maybe. Would this section be the "New to OCaml" section? Or some other? (Like "Melange for X users" in the JS section...)

docs/learn.md
jchavarri: Removed "problematic" and expanded a bit in https://github.com/melange-re/documentation-site/pull/12/commits/127995137166ab4cc9c597b6ff9266fa2c7865e2.

docs/learn.md
jchavarri: All the docs should be revisited with Reason syntax in mind. Some other things that comes to mind are:
- Pipe operator (`|.` vs `->`)
- Some syntax like extensions differ (e.g. two percentage signs in `ml` are one in `re`)

I would rather tackle all this as a focused task. Added some additional note in #7: https://github.com/melange-re/documentation-site/issues/7#issuecomment-1549519557

docs/learn.md
jchavarri: Removed in https://github.com/melange-re/documentation-site/pull/12/commits/ce220da061bc231ee5c1cb4f6ab96a2ce0167746 (also replaced the `…​` weird char with `...`)

docs/learn.md
jchavarri: There is a "Chaining" section before this one that talks about it, but I agree better to remove it to avoid distractions. Done in https://github.com/melange-re/documentation-site/pull/12/commits/ce220da061bc231ee5c1cb4f6ab96a2ce0167746.

jchavarri:
@anmonteiro thanks for the reviews! I think I went through all the comments and suggestions. I am going to merge this PR as it grew quite large, and GH UI makes it hard to deal with inline comments due to collapsing etc. But if there is anything missing or that needs improvement, please comment and I will tackle in another branch.

docs/learn.md
anmonteiro: yup, section looks good.

docs/learn.md
anmonteiro: Not a bad idea, and also something we can do anytime. no rush.

PR #23, opened 2023-05-16
add `bs.inline` section

### LINKED ISSUE #15: Document `bs.inline`

anmonteiro:
fixes #15


docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
 
+## Inlining constant values
+
+Some JavaScript idioms require special constants to be inlined since they serve
+as de-facto directives for bundlers. A common example is `process.env.NODE_ENV`:
+
+```js
+if(process.env.NODE_ENV !== 'production') {
+   // Development-only code
+}
```
jchavarri, line 561: Suggesting same style and formatting as generated output for consistency.

```suggestion
if (process.env.NODE_ENV !== "production") {
  // Development-only code
}
```

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+```
+
+becomes:
+
+```js
+if('development' !== 'production') {
+   // Development-only code
+}
```
jchavarri, line 569:
```suggestion
if ("development" !== "production") {
  // Development-only code
}
```

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+```
+
+In this case, bundlers such as Webpack can tell that the `if` statement always
+evaluates to a specific branch and eliminate the others.
+
+Melange provides the `[@bs.inline]` construct to achieve the same goal in
```
jchavarri, line 575: I have been trying (not sure if I was consistent enough) to refer to attributes by just their name, without the syntactic `[]` brackets. Wdyt?

```suggestion
Melange provides the `bs.inline` attribute to achieve the same goal in
```

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+generated JavaScript. Let's look at an example:
```
jchavarri, line 576:
```suggestion
generated JavaScript. Let’s look at an example:
```

https://english.stackexchange.com/a/36048

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+
+```ocaml
+external node_env : string = "NODE_ENV" [@@bs.val] [@@bs.scope "process", "env"]
+
+(* This variable will be used in the comparison below. *)
```
jchavarri, line 581: Is this comment necessary? The value is used in the line below.

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+let development = "development"
+let () = if node_env <> development then Js.log "Only in Production"
+
+(* The value of this variable will be inlined in the final output. The variable
+   won't be present in the JavaScript output. *)
```
jchavarri, line 586: Also unsure about this comment, as it is mentioned in the "As we can see" paragraph below.

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+let development_inline = "development" [@@bs.inline]
+let () = if node_env <> development_inline then Js.log "Only in Production"
+```
+
+As we can see in the generated JavaScript presented below:
+
+- the `development` variable is emitted
+    - it gets used as a variable `process.env.NODE_ENV !== development` in the
+      `if` statement
+- the `development_inline` variable isn't present in the final output
```
jchavarri, line 596:
```suggestion
- the `development_inline` variable isn’t present in the final output
```

docs/communicate-with-javascript.md
```diff
@@ -550,6 +550,71 @@ let () = match [%bs.external __filename] with
...
+if (process.env.NODE_ENV !== development) {
+  console.log("Only in Production");
+}
+
+if (process.env.NODE_ENV !== "development") {
+  console.log("Only in Production");
+}
+
+exports.development = development;
+/*  Not a pure module */
```
jchavarri, line 615: We can skip the first lines and exports from the generated JS, to avoid some noise (I don't think they add much when presented in the docs):

```suggestion
var development = "development";

if (process.env.NODE_ENV !== development) {
  console.log("Only in Production");
}

if (process.env.NODE_ENV !== "development") {
  console.log("Only in Production");
}
```

docs/communicate-with-javascript.md
```diff
@@ -1044,7 +1109,7 @@ var Path = require("path");
 var v = Path.join("a", 2);
 ```
 
-### Bing to a polymorphic function
```
jchavarri, line 1047: 🤣 🙏

docs/communicate-with-javascript.md
```diff
@@ -1267,6 +1332,9 @@ compile the first argument of the JavaScript function to the string `"exit"`.
 You can also use any JSON literal by passing a quoted string to `bs.as`: `bs.as
 {json|true|json}` or `bs.as {json|{"name": "John"}|json}`.
 
+Sometimes, passing a constant argument to an external function is not enough.
+If you need to inline constant values, check [`[@bs.inline]`](#bs-inline)
```
jchavarri, line 1336: Same suggestion as above about brackets:

```suggestion
If you need to inline constant values, check the [`bs.inline`](#bs-inline) attribute.
```

docs/communicate-with-javascript.md
```diff
@@ -1267,6 +1332,9 @@ compile the first argument of the JavaScript function to the string `"exit"`.
 You can also use any JSON literal by passing a quoted string to `bs.as`: `bs.as
 {json|true|json}` or `bs.as {json|{"name": "John"}|json}`.
 
+Sometimes, passing a constant argument to an external function is not enough.
```
jchavarri, line 1335: "is not enough" for what? I am not sure I get the connection between this "constant values as arguments" section and the `bs.inline` attribute 🤔

docs/communicate-with-javascript.md
anmonteiro: It's about the generated code. Should we make it more explicit?

docs/communicate-with-javascript.md
anmonteiro: OK I'll remove the comments.

docs/communicate-with-javascript.md
anmonteiro: this connection probably doesn't make sense indeed. I'll remove.

anmonteiro:
I'm going to merge this, let me know if you have additional comments.

docs/communicate-with-javascript.md
jchavarri: I think it's Option + Shift + ] on macos us keyboards. In any case, I can make a find/replace when we are done with v1.

jchavarri:
Looks great ✨

PR #26, opened 2023-05-17
comm with js: add docs for `bs.scope`

### LINKED ISSUE #13: Document `bs.scope`

jchavarri:
Fixes #13.

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -339,8 +339,8 @@ These attributes are used to annotate arguments in `external` definitions:
   (automated)
 - [`bs.unwrap`](#approach-2-polymorphic-variant-bsunwrap): unwrap variant values
 
-These attributes are used in places like records, fields, parameters,
-functions, and more:
+These attributes are used in places like records, fields, parameters, functions,
+and more:
 
 - `bs.as`: redefine the name generated in the JavaScript output code. Used in
   [constant function arguments](#constant-values-as-arguments),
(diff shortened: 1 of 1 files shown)


docs/communicate-with-javascript.md
```diff
@@ -826,6 +841,91 @@ var Path = require("path");
...
+var ExpoCamera = require("expo-camera");
+
+var camera_type_back = ExpoCamera.Camera.Constants.Type.back;
+```
+
+It can also be used in combination with other attributes besides `bs.module`,
+like `bs.val`:
+
+```ocaml
+external imul : int -> int -> int = "imul" [@@bs.val] [@@bs.scope "Math"]
```
jchavarri, line 898: I realized `bs.scope` is not necessary when using `bs.val` because the external function name can take dots, e.g.

https://github.com/melange-re/documentation-site/blob/7022ed1542e940dde957bbde0eed501811418f76/docs/communicate-with-javascript.md?plain=1#L801

But as it's supported, why not document it 🤷

docs/communicate-with-javascript.md
```diff
@@ -339,7 +339,8 @@ These attributes are used to annotate arguments in `external` definitions:
   (automated)
 - [`bs.unwrap`](#approach-2-polymorphic-variant-bsunwrap): unwrap variant values
 
-These attributes are used in other places like records, fields, parameters, functions...:
+These attributes are used in other places like records, fields, parameters,
+functions...:
```
anmonteiro, line 343:
```suggestion
These attributes are used in places like records, fields, parameters,
functions, and more:
```

docs/communicate-with-javascript.md
anmonteiro: I don't know if we want to rely on that. I could imagine some transformations we would want to do in the future would benefit from having well-defined semantics.

jchavarri: I agree. This made me think we should then get rid of the `bs.val` example what was using `"Math.imul"` string on the right side of the external. Did so in https://github.com/melange-re/documentation-site/pull/26/commits/8f8cdda0767d67505679d198e4f2aaec81480909.

jchavarri: > But as it's supported, why not document it 🤷

Oh, just read this 😬 Well, lmk if you want to change it back. I still think if it's supported but we don't want people to rely much on it, and there's a "safer" alternative to it, better to leave it undocumented.

PR #28, opened 2023-05-18
comm with js: code generation / `bs.deriving`

### LINKED ISSUE #16: Document `bs.optional`

jchavarri:
Fixes #16 and #14.

I had a lot of trouble coming up with a structure that made sense for this section. In the end, I decided to arrange the information based on the types that can be used with `bs.deriving`. This made it easier to arrange the possibilities.

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -200,7 +200,7 @@ them before doing so.
 
 - Variants and polymorphic variants: Better transform them into readable
   JavaScript values before manipulating them from JavaScript, Melange provides
-  [some helpers](todo-fix-me.md) to do so.
+  [some helpers](#deriving-code-from-types) to do so.
 - Exceptions
 - Option (a variant type): Better use the `Js.Nullable.fromOption` and
   `Js.Nullable.toOption` functions in the [`Js.Nullable` module](todo-fix-me.md)
@@ -309,7 +309,7 @@ These attributes are used to annotate `external` definitions:
 - [`bs.module`](#using-functions-from-other-javascript-modules): bind to a value
   from a JavaScript module
 - [`bs.new`](#javascript-classes): bind to a JavaScript class constructor
-- [`bs.obj`](#using-jst-objects): create JavaScript object
+- [`bs.obj`](todo-fix-me.md): create JavaScript object
 - [`bs.return`](#wrapping-returned-nullable-values): automate conversion from
   nullable values to `Option.t` values
 - [`bs.send`](#calling-an-object-method): call a JavaScript object method using
@@ -339,7 +339,7 @@ These attributes are used to annotate arguments in `external` definitions:
   (automated)
 - [`bs.unwrap`](#approach-2-polymorphic-variant-bsunwrap): unwrap variant values
 
-These attributes are used in places like records, fields, parameters, functions,
+These attributes are used in places like records, fields, arguments, functions,
 and more:
 
 - `bs.as`: redefine the name generated in the JavaScript output code. Used in
@@ -371,7 +371,7 @@ Here is the list of all extensions supported by Melange:
 
 - `bs.debugger`: insert `debugger` statements
 - `bs.external`: read global values
-- `bs.obj`: create JavaScript object literals
+- [`bs.obj`](#using-jst-objects): create JavaScript object literals
 - `bs.raw`: write raw JavaScript code
 - `bs.re`: insert regular expressions
 
@@ -573,8 +573,8 @@ if ("development" !== "production") {
 In this case, bundlers such as Webpack can tell that the `if` statement always
 evaluates to a specific branch and eliminate the others.
 
-Melange provides the `bs.inline` attribute to achieve the same goal in
-generated JavaScript. Let’s look at an example:
+Melange provides the `bs.inline` attribute to achieve the same goal in generated
+JavaScript. Let’s look at an example:
 
 ```ocaml
 external node_env : string = "NODE_ENV" [@@bs.val] [@@bs.scope "process", "env"]
@@ -592,7 +592,8 @@ As we can see in the generated JavaScript presented below:
     - it gets used as a variable `process.env.NODE_ENV !== development` in the
       `if` statement
 - the `development_inline` variable isn’t present in the final output
-    - its value is inlined in the `if` statement: `process.env.NODE_ENV !== "development"`
+    - its value is inlined in the `if` statement: `process.env.NODE_ENV !==
+      "development"`
 
 ```js
 var development = "development";
@@ -1294,7 +1295,7 @@ Some JavaScript APIs take a limited subset of values as input. For example,
 Node’s `fs.readFileSync` second argument can only take a few given string
 values: `"ascii"`, `"utf8"`, etc. Some other functions can take values from a
 few given integers, like the `createStatusBarItem` function in VS Code API,
-which can take an alignment parameter that can only be [`1` or
+which can take an `alignment` parameter that can only be [`1` or
 `2`](https://github.com/Microsoft/vscode/blob/2362ec665c84a1519162b50c36ed4f29d1e20f62/src/vs/vscode.d.ts#L4098-L4109).
 
 One could still type these parameters as just `string` or `int`, but this would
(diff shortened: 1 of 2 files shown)


docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+  age : int;
+}
+[@@bs.deriving abstract]
+```
+
+The accessors `nameGet` and `ageGet` will still be generated, but not the
+constructor `person`. This is useful when binding to JavaScript objects while
+preventing any Melange code from creating values of such type.
+
+### Create JavaScript objects using external functions
```
jchavarri, line 2077: I am not sure where this section should be. It is not `bs.deriving` but it is kind of "code generation". One alternative would be to move it to "Bind to JS objects" section... 🤔

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+console.log(Belt_Array.map(pets, name).join("&"));
+```
+
+#### Convert records into abstract types
+
+Usually, it’s recommended to use plain records when compiling to JavaScript
+objects (see the section on [binding to JavaScript
+objects](#bind-to-javascript-objects)). But there's a specific case where
+records may not be enough. This happens when a record type has optional fields,
+and we want the resulting JavaScript objects to exclude those fields entirely.
```
anmonteiro, line 1896:
```suggestion
records may not be enough: Specifically, when we want to emit a JavaScript object
where some of the keys may be nullable, i.e. not present at all in the JavaScript
object. `
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+```ocaml
+type person = {
+  name : string;
+  age : int option;
+}
+```
+
+The specific use-case we are referring to arises when we have a Melange value
+such as `{ name = "John"; age = None }`, but we desire the resulting JavaScript
+object to be `{name: "Carl"}`.
```
anmonteiro, line 1909:
```suggestion
An example of this use-case would be expecting `{ name = "John"; age = None }`
to generate a JavaScript such as `{name: "Carl"}`, where the `age` key doesn't
appear.
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+
+To address this case, the `@bs.deriving abstract` attribute can prove useful.
+When applied to a record type, this attribute removes the record type definition
+and replaces it with an abstract type. Additionally, it generates these
+functions:
```
anmonteiro, line 1914:
```suggestion
The `@bs.deriving abstract` attribute exists to solve this problem. When
present in a record type, `@bs.deriving abstract` makes the record definition
abstract and generates the following functions instead:
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+
+- A constructor function for creating values of the type
+- Getters and setters for accessing the record fields
+
+Let’s see an example. Considering this Melange code:
```
anmonteiro, line 1919:
```suggestion
`@bs.deriving abstract` effectively models a record-shaped JavaScript object
exclusively through a set of (generated) functions derived from attribute annotations
on the OCaml type definition.

Let’s see an example. Considering this Melange code:
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+type person = {
+  name : string;
+  age : int; [@bs.optional]
+}
+[@@bs.deriving abstract]
+```
+
+Melange will transform the `person` type into an abstract type and generate the
+constructor, getters, and setters functions. So we would get an interface like
+this:
```
anmonteiro, line 1931:
```suggestion
Melange will make the `person` type abstract and generate constructor, getter
and setter functions. In our example, the OCaml signature would look like this after
preprocessing:
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+
+val nameGet : person -> string
+
+val ageGet : person -> int option
+```
+
+The `person` constructor enables the creation of values for the `person` type.
+It is the only valid way to create such values because using literals directly
+like `{ name = "Alice"; age = None }` will no longer type check, due to the
+opaque nature of the abstract type definition.
```
anmonteiro, line 1946:
```suggestion
The `person` function can be used to create values of `person`. It is the only
possible way to create values of this type, since Melange makes it abstract.
Using literals like `{ name = "Alice"; age = None }` directly doesn't type check.
```

docs/communicate-with-javascript.md
anmonteiro: I'd probably move it to the "Bind to JS objects" section, indeed.

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+[@@bs.deriving abstract]
+```
+
+The accessors `nameGet` and `ageGet` will still be generated, but not the
+constructor `person`. This is useful when binding to JavaScript objects while
+preventing any Melange code from creating values of such type.
+
+### Create JavaScript objects using external functions
+
+We have already explored one approach to creating JavaScript object literals by
```
anmonteiro, line 2079:
```suggestion
We have already explored one approach for creating JavaScript object literals by
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+using [`Js.t` values and the `bs.obj` extension](#using-jst-objects).
+
+In addition to that approach, Melange offers the `bs.obj` attribute, which can
```
anmonteiro, line 2082:
```suggestion
Melange additionally offers the `bs.obj` attribute, which can
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+be used in combination with external functions to create JavaScript objects.
+When these functions are called, they generate objects with fields corresponding
+to the labeled arguments of the function.
+
+If any of these labeled arguments are defined as optional and omitted during
+function application, the resulting JavaScript object will exclude the
+corresponding fields. This feature allows for runtime object creation with
+dynamic selection of the subset of fields to include.
```
anmonteiro, line 2090:
```suggestion
corresponding fields. This allows to create runtime objects and control whether
optional keys are emitted at runtime.
```

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+syntactically valid. The value of this string is ignored by the compiler.
+
+Since there is an optional argument `options`, an additional unlabeled argument
+of type `unit` is required. This allows the compiler to determine if the
+optional argument has been omitted when the function is applied. More
+information about optional arguments can be found in the [OCaml
+manual](https://v2.ocaml.org/manual/lablexamples.html#s:optional-arguments).
+
+The return type of the function should be left unspecified using the wildcard
+type `_`. Melange will automatically infer the type of the resulting JavaScript
```
anmonteiro, line 2127: I don't like this recommendation. At the very least it should be `< .. > Js.t`, which makes it a lot more explicit what's going on.

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+  unit ->
+  _ = ""
+  [@@bs.obj]
+```
+
+Note that the empty string at the end of the function is used to make it
+syntactically valid. The value of this string is ignored by the compiler.
+
+Since there is an optional argument `options`, an additional unlabeled argument
+of type `unit` is required. This allows the compiler to determine if the
```
anmonteiro, line 2121: The argument isn't really required, it's just a convenience that allows not having to apply every optional argument with `?foo:None`.

Feel free to leave it as is, though? it's a good recommendation, it's just not "required".

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+binding to JavaScript objects with fields that are reserved keywords in OCaml,
+Melange allows the use of an underscore prefix for the labeled arguments. The
+resulting JavaScript object will have the underscore removed from the field
+names. This is only required for the `bs.obj` attribute, while for other cases,
+the `bs.as` attribute can be used to rename fields.
+
+If we call the function like this:
+
+```ocaml
+let homeRoute = route ~__type:"GET" ~path:"/" ~action:(fun _ -> Js.log "Home") ()
```
anmonteiro, line 2140: There's an extra underscore in `~__type`, is it intentional? I didn't check how it renders.

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+```
+
+We would get the generated JavaScript, which does not include the `options`
+field as required:
```
anmonteiro, line 2144:
```suggestion
We get the following JavaScript, which does not include the `options`
field since its argument wasn't present:
```

anmonteiro (commented):
The structure looks good overall, but I'd reword some things. I left suggestions as appropriate.

I think it's important that the `bs.deriving abstract` section also mentions `bs.as`, and it currently doesn't.

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
 
 `identity` will make sure that compiler will do nothing about the returned
 value. It is rarely used, but introduced here for debugging purposes.
+
+## Deriving code from types
+
+As we saw in a [previous section](#non-shared-data-types), there are some types
+in Melange that compile to values that are not easily manipulable from
```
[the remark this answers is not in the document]
jchavarri: I don't think the suggested sentence is grammatically correct.

docs/communicate-with-javascript.md
jchavarri: https://github.com/melange-re/documentation-site/pull/28/commits/987495dc2496d926b22108832a2b38a342511b1f.

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
 
 `identity` will make sure that compiler will do nothing about the returned
 value. It is rarely used, but introduced here for debugging purposes.
+
+## Deriving code from types
```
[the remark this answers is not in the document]
jchavarri: Renamed to "Generate getters, setters and constructors" in https://github.com/melange-re/documentation-site/pull/28/commits/924286e9df03305dc924c82756a71b2e74907365

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+#### Conversion functions
+
+Use `@bs.deriving jsConverter` on a variant type to create converter functions
+that allow to transform back and forth between JavaScript integers and Melange
+variant values.
+
+There are a few differences with `@bs.deriving accessors`:
+
+- `jsConverter` works with the `bs.as` attribute, while `accessors` does not
+- `jsConverter` does not support variant tags with payload, while `accessors`
```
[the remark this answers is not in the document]
jchavarri: Do you want me to create an issue on melange-re/melange?

docs/communicate-with-javascript.md
```diff
@@ -1669,3 +1670,485 @@ nullable]` above. Currently 4 directives are supported: `null_to_opt`,
...
+  }
+];
+
+console.log(Belt_Array.map(pets, name).join("&"));
+```
+
+#### Convert records into abstract types
+
+Usually, it’s recommended to use plain records when compiling to JavaScript
+objects (see the section on [binding to JavaScript
```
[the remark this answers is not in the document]
jchavarri: tried something in https://github.com/melange-re/documentation-site/pull/28/commits/a1d961d6d653d182b49d2126d785a1fc0fce589b.

docs/communicate-with-javascript.md
jchavarri: It does not matter really. We can change the recommendation, but if you write 

```
external route :
  _type:string ->
  path:string ->
  action:(string list -> unit) ->
  ?options:< .. > ->
  unit ->
  int = ""
  [@@bs.obj]
```

`melc` will still compile just fine.

jchavarri: So it seems a bit useless, as a user, to type more characters that do not really affect the outcome in any sense. Do you mean to recommend using `< .. > Js.t` as some sort of documentation?

docs/communicate-with-javascript.md
jchavarri: Tried to rewrite in https://github.com/melange-re/documentation-site/pull/28/commits/04ef00465e6834d0d4bf6fbb990af8e25c51ef44.

docs/communicate-with-javascript.md
jchavarri: No, was a typo. Fixed in https://github.com/melange-re/documentation-site/pull/28/commits/584f30bb1af73a3921ab32dd4b28b0e443ec148a.

jchavarri:
> I think it's important that the `bs.deriving abstract` section also mentions `bs.as`, and it currently doesn't.

I tried to make `bs.as` work with `bs.deriving abstract` but was not able to, I get the `Error (warning 101 [unused-bs-attributes]): Unused attribute: bs.as` error. I could only make it work with `jsConverter`.

Could you provide a working example with `bs.deriving abstract`  and `bs.as`?

jchavarri:
I added a note that `bs.as` can be used with records and `bs.deriving` but didn't think it was necessary to add an example: https://github.com/melange-re/documentation-site/pull/28/commits/cdd06a3fd95cc986bc6a4d308a1b53529afe17c3

Gonna go ahead and merge, but lmk if anything can be improved.

PR #32, opened 2023-05-19
comm with js: add "language concepts" section

jchavarri:
This PR reorganizes some of the content in "Communicate with JavaScript" page. It creates a new section "Language concepts" that is placed at the top of the page.

This new section serves two goals:

- Make the index of the page a bit cleaner, by grouping under a single section the content that is more about underlying concepts, and is not part of the "how-to guides" that come after
- Make room for the soon-to-be-written [pipe operator docs](#19), which will be included in this new "Language concept" section

For now, the structure of the new "Language concepts" section looks like:

- Attributes and extension nodes
- External functions (renamed from "Foreign function interface" for simplicity)
  - Special identity external
- Abstract types

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -2,10 +2,179 @@
 
 Melange interoperates very well with JavaScript, and provides a wide array of
 features to consume foreign JavaScript code. To learn about these techniques, we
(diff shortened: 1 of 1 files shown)


anmonteiro (approved):
nice, I like this new section

PR #35, opened 2023-05-19
comm with js: unboxed

### LINKED ISSUE #17: Document OCaml `unboxed` attribute for Melange

jchavarri:
Fixes #17.

Based on #32 (`comm-with-js/lang-concepts`).

I decided to mention the `unboxed` attribute as part of the "Language concepts" section that talks about attributes. Hopefully makes sense this way. I think it's a quite niche optimization that does not deserve its own section, and keeping it in the attributes section allowed to mention also other pre-built attrs like `alert` and `deprecated` that Melange programs can use as well.

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -72,9 +72,12 @@ The difference between one and two percentage characters is detailed in the
 
 Attributes are "decorations" applied to specific parts of the code to provide
 additional information. In Melange, attributes are used in two ways to enhance
-the expressiveness of generating JavaScript code.
+the expressiveness of generating JavaScript code: either reusing existing OCaml
+built-in attributes or defining new ones.
 
-The first approach is leveraging the existing [OCaml's built-in
+##### Reusing OCaml attributes
+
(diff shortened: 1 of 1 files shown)


docs/communicate-with-javascript.md
```diff
@@ -63,31 +65,60 @@ var a = 1; var b = 2
...
-JavaScript code:
+additional information. In Melange, attributes are used in two ways to enhance
+the expressiveness of generating JavaScript code.
 
-```ocaml
-type t = {
-  foo : int; [@bs.as "foo_for_js"]
-  bar : string;
-}
+The first approach is leveraging the existing [OCaml's built-in
```
anmonteiro, line 77: you describe the "what" but not the "why". I think we should at least include a tiny sentence that says it's useful for heterogeneous collections or something.

jchavarri: added in https://github.com/melange-re/documentation-site/pull/35/commits/d9d6c637fc410167d553d33652e3e84b223d3409.

PR #37, opened 2023-05-22
Reason syntax: explore cmarkit

### LINKED ISSUE #7: Show Reason syntax

jchavarri:
Fixes #7.

This PR explores the idea of using cmarkit to process markdown files and automatically generate Reason code blocks from the OCaml ones.

The PR includes a small executable that will read from stdin, assuming is the string from a given markdown page of the site, and then will:
- analyze it to find `ocaml` code blocks
- pass them through `Refmt` functions to get `reason` syntax
- output the whole mapped file on stdout

It also defines a dune alias and one dune rule that process 1 page as demo, some other things would need to be done if we decide to move forward with this approach:

### TODO 

- [x] Process more pages besides "Comm with JS" (maybe can use some generic dune rule with `*.md`)
- [x] Handle the case where the `reason` blocks exist already
- [x] On top of the above, would be nice to have a "check" mode that doesn't write anything, just verifies the snippets are correct. Useful in CI
- [x] Introduce a UI switch to hide/show each syntax code blocks based on user needs
- [x] Investigate if it's possible to have dune to promote the file with the same name as the original source.


jchavarri:
PR is ready. In addition to the items mentioned in description, I pinned to latest cmarkit to benefit from a fix for some issue with code spans inside tables.

All places that use OCaml snippets are now switched to the Reason version when clicking on the top left button that also was added (see images below).

The only place where Reason syntax remains "hard-coded" and can't be switched to OCaml is in the "Melange for X devs" section. I think it makes sense because in that section we are promoting Reason to the syntax that we want to use.

imagen
imagen

jchavarri:
Gonna merge as it's mostly auto-generated content, happy to tackle any change reqs in a new PR.

PR #56, opened 2023-05-26
Fixed instructions on building app after an alias has been created

psb:
If the reader has been going through the example code and created an alias in the earlier section 'Using aliases', they need to build the project (in the 'Handling assets' section) using the alias `my-app`.


jchavarri:
Thanks!

PR #57, opened 2023-05-26
comm with js: add exports

### LINKED ISSUE #18: Document "Export to JS" and `default` exports

jchavarri:
Fixes #18.

- Adds some docs about `module_systems` field in "Build system"
- Adds a section "Use Melange code from JavaScript" in "Comm with JS"

### DIFF
--- a/docs/build-system.md
+++ b/docs/build-system.md
@@ -436,8 +436,8 @@ to configure it to generate
 [ES6](https://en.wikipedia.org/wiki/ECMAScript#6th_Edition_-_ECMAScript_2015)
 modules.
 
-The place to do so is the `melange.emit` stanza. To configure Melange to
-generate ES6 modules, we will use the `module_systems` field:
+Use the `module_systems` field in the `melange.emit` stanza to emit
+ES6 modules:
 
 ```text
 (melange.emit
--- a/docs/build-system.md
+++ b/docs/build-system.md
@@ -169,7 +169,8 @@ let () = Js.log(Lib.name);
 
 The `melange.emit` stanza tells Dune to generate JavaScript files from a set of
 libraries and modules. In-depth documentation about this stanza can be found in
-the [Dune docs](https://dune.readthedocs.io/en/stable/melange.html).
+the [Dune
+docs](https://dune.readthedocs.io/en/stable/melange.html#melange-emit).
 
 The file structure of the app should look something like this:
 
@@ -346,8 +347,8 @@ provides a way to specify these dependencies, depending on the stanza:
 - For `melange.emit` stanzas, a field `runtime_deps`
 
 Both fields are documented in the [Melange
-page](https://dune.readthedocs.io/en/stable/melange.html) of the Dune
-documentation site.
+page](https://dune.readthedocs.io/en/stable/melange.html#melange-emit) of the
+Dune documentation site.
 
 For the sake of learning how to work with assets in a Melange project, let’s say
 that we want to read the string in `Lib.name` from a text file. We will combine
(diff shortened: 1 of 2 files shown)


docs/build-system.md
```diff
@@ -423,3 +424,33 @@ Melange developers. For further details about how Dune works and its integration
...
+declare imports for the values and modules they depend on.
+
+By default, Melange will produce
+[CommonJS](https://en.wikipedia.org/wiki/CommonJS) modules, but it is possible
+to configure it to generate
+[ES6](https://en.wikipedia.org/wiki/ECMAScript#6th_Edition_-_ECMAScript_2015)
+modules.
+
+The place to do so is the `melange.emit` stanza. To configure Melange to
+generate ES6 modules, we will use the `module_systems` field:
```
anmonteiro, line 440:
```suggestion
Use the `module_systems` field in the `melange.emit` stanza to emit
ES6 modules:
```

docs/build-system.md
```diff
@@ -423,3 +424,33 @@ Melange developers. For further details about how Dune works and its integration
...
+ (module_systems es6))
+```
+
+If no extension is specified, the resulting JavaScript files will use `.js`. You
+can specify a different extension with a pair `(<module_system> <extension>)`,
+e.g. `(module_systems (es6 mjs))`. Multiple module systems can be used in the
+same field as long as their extensions are different. For example,
+`(module_systems commonjs (es6 mjs))` will produce one set of JavaScript files
+using CommonJS and the `.js` extension, and another using ES6 and the `.mjs`
+extension.
```
anmonteiro, line 456: should we link to https://dune.readthedocs.io/en/stable/melange.html#melange-emit where this is also documented?

docs/communicate-with-javascript.md
```diff
@@ -3029,3 +3029,77 @@ type person =
 The accessors `nameGet` and `ageGet` will still be generated, but not the
 constructor `person`. This is useful when binding to JavaScript objects while
 preventing any Melange code from creating values of such type.
+
+## Use Melange code from JavaScript
+
+As mentioned in the [build system
+section](build-system.md#commonjs-or-es6-modules), Melange allows to produce
+both CommonJS and ES6 modules. In any of both cases, there is nothing special to
```
anmonteiro, line 3037:
```suggestion
both CommonJS and ES6 modules. In both cases, there is nothing special to
```

docs/communicate-with-javascript.md
```diff
@@ -3029,3 +3029,77 @@ type person =
...
+do to use your Melange-generated JavaScript code from any hand-written
```
anmonteiro, line 3038: instead of "nothing special to do" can we mention "it works as expected"?

docs/communicate-with-javascript.md
```diff
@@ -3029,3 +3029,77 @@ type person =
...
+function print(name) {
+  return "Hello" + name;
+}
+
+export {
+  print ,
+}
+```
+
+So one can use either `require` or `import` (depending on the modules system of
```
anmonteiro, line 3073:
```suggestion
So one can use either `require` or `import` (depending on the module system of
```

docs/build-system.md
jchavarri: Added in https://github.com/melange-re/melange-re.github.io/pull/57/commits/89610d28ab69529dcfb70659bdfdb5d882e3121b (we link to these docs 3 times now from this page 😄 but it doesn't hurt I guess)

docs/communicate-with-javascript.md
jchavarri: https://github.com/melange-re/melange-re.github.io/pull/57/commits/945a891bc6b9e67661b34fd5c6cd165d07677b0b

anmonteiro (approved):

PR #70, opened 2023-05-31
bs.splice to bs.variadic

psb:
Changed bs.splice to bs.variadic and added note about bs.variadic replacing bs.splice.

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -739,7 +739,7 @@ These attributes are used to annotate `external` definitions:
 - [`bs.val`](#bind-to-global-javascript-functions-or-values): bind to global
   JavaScript functions or other values
 - [`bs.variadic`](#variadic-function-arguments): bind to a function taking
-  variadic arguments from an array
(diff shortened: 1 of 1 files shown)


docs/communicate-with-javascript.md
```diff
@@ -739,7 +739,7 @@ These attributes are used to annotate `external` definitions:
 - [`bs.val`](#bind-to-global-javascript-functions-or-values): bind to global
   JavaScript functions or other values
 - [`bs.variadic`](#variadic-function-arguments): bind to a function taking
-  variadic arguments from an array
+  variadic arguments from an array (`bs.variadic` has replaced `bs.splice`, which is now considered deprecated; `bs.splice` still exists and can be used for backwards compatibility)
```
jchavarri, line 742: As the attributes are ordered alphabetically for easier finding, let's leave the variadic entry as it was, and add an new entry for `bs.splice` above, after `bs.scope`:

```
- [`bs.splice`](#variadic-function-arguments): a deprecated attribute, is an
  alternate form of `bs.variadic`
```

jchavarri (commented):
Looks good, I think we can add an entry for `splice` attribute.

docs/communicate-with-javascript.md
psb: Okay. I've updated the text.

PR #72, opened 2023-06-01
Publish versions

### LINKED ISSUE #47: Support some kind of versioning system
From @anmonteiro:

> we should consider publishing the docs to a path like `/stable` so that we can keep newer branches as we document stuff for dev features

jchavarri:
Fixes #47.

Uses [mike](https://github.com/jimporter/mike) to achieve so. The process is quite simple, `mike` handles folders for each version in the `gh-pages` branch.

- Pushes to `master` deploy to the `unstable` version
- A GH action is configured so that new tags with the `v*` format deploy a new version with the tag name, aliased to `stable`.

### DIFF
--- a/theme/CNAME
+++ b/theme/CNAME
@@ -1,1 +1,0 @@
-melange.re
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,4 +15,8 @@ jobs:
         with:
           python-version: 3.x
       - run: python3 -m pip install -r ./pip-requirements.txt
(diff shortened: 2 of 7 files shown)


theme/CNAME
```diff
@@ -1 +0,0 @@
-melange.re
```
jchavarri, line 1: This are not copied by `mike` so added them in the `gh-pages` branch manually: https://github.com/melange-re/melange-re.github.io/commit/42b5b09021041cc312f389afc770f7a5b82c1389

PR #73, opened 2023-06-01
Updated 'Wrapping returned nullable values' with ReasonML fixes.

psb:


docs/communicate-with-javascript.md
```diff
@@ -2430,10 +2430,10 @@ Note that the first argument will be reserved for `this`.
 ### Wrapping returned nullable values
 
 JavaScript models `null` and `undefined` differently, whereas it can be useful
-to treat both as `'a option` in OCaml.
+to treat both as <span class="text-ocaml">`'a option` in OCaml</span><span class="text-reasonml">`option('a)` in Reason</span>.
```
jchavarri, line 2433: Let's use "in Melange":

```suggestion
to treat both as <span class="text-ocaml">`'a option`</span><span class="text-reasonml">`option('a)`</span> in Melange.
```

jchavarri (commented):
Great catch, maybe we can simplify a bit.

jchavarri:
Thanks!

PR #83, opened 2023-06-14
Added docs for finding and using packages.

psb:


docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+   bs-json
+   (subdir
+    src
+    (library
+     (name bs_json)
+     (wrapped false)
+     (modes melange))))))
+```
+
+(If the `dune` file contains the line `(dirs :standard \ node_modules)`, it should be removed.)
```
jchavarri, line 379: No need for parenthesis.

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (module_systems es6))
+```
+
+Depending on where the installed package is used will determine in which `dune` file the package should be referenced. In the next section we will see how to reference a package from within a library subfolder.
+
+### NPM packages
+
+A number of Melanage compatible packages can be found on NPM. These packages are typically bindings to JavaScript libraries. Many older, but still useful, compatible BuckleScript libraries can be found on NPM, e.g., `bs-json`. An NPM package is installed like any normal NPM package is: `npm install @glennsl/bs-json`, which will add the package to the `package.json` file at the root of our project.
+
+In order to get dune to pick up and process the NPM package so that it can be consumed in our application, the `dune` file at the root of our project must be updated with a `subdir` stanza:
```
jchavarri, line 360: Let's use npm in lower case, as it shows on their website (and other pages of the melange docs), and Dune with upper case:

```suggestion
In order to get Dune to pick up and process the npm package so that it can be consumed in our application, the `dune` file at the root of our project must be updated with a `subdir` stanza:
```

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+In our project structure above we have the file <code class="text-ocaml">data.ml</code><code class="text-reasonml">data.re</code> under the folder `src/lib`. If we want use the `bs-json` package from within the <code class="text-ocaml">data.ml</code><code class="text-reasonml">data.re</code> file then we need to add the package name to the `dune` file in the same folder, i.e., `src/lib/dune`:
+
+```text
+(library
+ (name data)
+ (libraries bs_json)
+ (modes melange))
+```
+
+Note that the package `bs-json` was renamed to `bs_json` in the `subdir` stanza and is referenced as `bs_json` in the `dune` file.
```
jchavarri, line 391: Again this seems to be conflating packages with libraries, but these are two different concepts. A package can install multiple libraries.

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+If we install any more NPM packages then a new `subdir` stanza will have to be added to `dune` file in the root of our project and referenced in the `dune` file that lives next the code we want to use the package from. See [this dune file](https://github.com/jchavarri/pupilfirst/blob/b6fe6fa27814336b9e92f965449f3e010e148c4b/dune) for a larger example that uses multiple NPM packages.
+
+**The advantage of publishing packages on opam is that none of the these `subdir` and `vendored_dirs` configurations are necessary.**
```
jchavarri, line 395: It is fine to include these docs for now as we transition packages to opam. But we should try to remove this section asap cc @davesnx @anmonteiro

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+```text
+...
+depends: [
+  ...
+  "reason-react" {>= "0.11.0"}
+]
+```
+
+To use the installed package, add the package name to the `dune` file under the `libraries` field. For example, if our project structure looks like:
```
jchavarri, line 284: Packages <> Libraries:

```suggestion
To use a library from the installed package, add the library name to the `dune` file under the `libraries` field. For example, if our project structure looks like:
```

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+(melange.emit
+ (target output)
+ (alias react)
+ (libraries melange lib reason-react)
+ (preprocess
+  (pps reactjs-jsx-ppx))
+ (module_systems es6))
+```
+
+(Some packages, like `reason-react`, will need to be preprocessed using a ppx, which may also need to be installed via opam.)
```
jchavarri, line 325:
```suggestion
Some libraries, like `reason-react`, work well with PPX rewriters like `reactjs-jsx-ppx`. These preprocessors may be installed together with the library as part of the same package, or they might be part of a different package, in which case they need to be installed separately.
```

jchavarri (commented):
Thanks! It's going in the right direction, but there is some confusion about packages and libraries. I left a few comments that go more in detail.

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+(melange.emit
+ (target output)
+ (alias react)
+ (libraries melange reason-react melange-fetch)
+ (preprocess
+  (pps reactjs-jsx-ppx))
+ (module_systems es6))
+```
+
+Depending on where the installed package is used will determine in which `dune` file the package should be referenced. In the next section we will see how to reference a package from within a library subfolder.
```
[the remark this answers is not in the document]
psb: Based on your corrections earlier I think "installed package" should be "installed library". With this sentence I am trying to explain that if I have `src/my_lib/my_library_file.re` then the installed library, e.g., bs-json, should be added to `src/my_lib/dune` and not in `src/dune` if I want to use bs-json in `src/my_lib/my_library_file.re`.

Would this be correct?

docs/package-management.md
psb: Which parens?

docs/package-management.md
psb: Yeah, I wasn't too sure of the difference between a package and library. I will change it to library.

docs/package-management.md
psb: Can reason-react work without the PPX? The term "work well" would seem to suggest that the reason-react library (got it right this time 😁 ) would still work without the PPX, just badly.

docs/package-management.md
jchavarri:
```suggestion
If the `dune` file contains the line `(dirs :standard \ node_modules)`, it should be removed.
```

docs/package-management.md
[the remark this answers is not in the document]
psb: I have deleted the paragraph as it is essentially repeated further down in the bs-json installation example.

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (name data)
+ (libraries bs_json)
+ (modes melange))
+```
+
+Note that the library `bs-json` was renamed to `bs_json` in the `subdir` stanza and is referenced as `bs_json` in the `dune` file.
+
+If we install any more npm packages then a new `subdir` stanza will have to be added to the `dune` file in the root of our project. See [this dune file](https://github.com/jchavarri/pupilfirst/blob/b6fe6fa27814336b9e92f965449f3e010e148c4b/dune) for a larger example that uses multiple npm packages.
+
+**The advantage of publishing packages on opam is that none of the these `subdir` and `vendored_dirs` configurations are necessary.**
```
anmonteiro, line 393: the other advantage is that you only have to build them once, (i.e. if you vendor libraries you need to build them as part of your project)

docs/package-management.md
```diff
@@ -267,3 +267,129 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
 inside the bindings— is that it gives users of the bindings complete flexibility
 over the way these JavaScript packages are downloaded and bundled.
 
+## Finding and using Melange compatible packages
```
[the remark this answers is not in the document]
anmonteiro: we should definitely recommend dune-release. It works well and is part of the platform

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+### Opam packages
```
anmonteiro, line 272:
```suggestion
### OPAM packages
```

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+Ideally, Melange compatible packages should be published on opam. To search for a package on opam enter `opam search <package_name>`, e.g., `opam search reason-react`. If the package is found, install it by running `opam install reason-react`. As mentioned earlier, opam will not add the package to the `.opam` file, so remember to add the package after it is installed:
```
anmonteiro, line 274:
```suggestion
Melange packages are usually available on [OPAM](https://opam.ocaml.org). Package search is available in the OPAM CLI via `opam search <package-name>`, e.g. `opam search reason-react`. You can run `opam install <package-name>` to download, build and install OPAM packages in your switch. Remember that OPAM won't automatically add the dependency to `<your-project>.opam` file, so it must be added manually:
```

docs/package-management.md
anmonteiro: Good point. I'd suggest "Some libraries will only work after being processed by an accompanying PPX"

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (alias react)
+ (libraries lib reason-react)
+ (preprocess
+  (pps reactjs-jsx-ppx))
+ (module_systems es6))
+```
+
+(Some packages, like `reason-react`, will need to be preprocessed using a ppx, which may also need to be installed via opam.)
+
+### Unpublished opam packages
```
anmonteiro, line 327: Let's uppercase OPAM (please check other places I might have missed)

```suggestion
### Unpublished OPAM packages
```

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+Some packages may have an `.opam` file, but they are not yet published in the official opam repository. These packages should be installed using `opam pin`, e.g., `opam pin add melange-fetch.dev https://github.com/melange-community/melange-fetch`. The project `.opam` file should then be updated in two places:
```
anmonteiro, line 329:
```suggestion
OPAM packages that have not yet been published may be installed with the `opam pin` command. For example, `opam pin add melange-fetch.dev git+https://github.com/melange-community/melange-fetch` will obtain `melange-fetch` from its Git repository and install it on your switch. Your `<your-project>.opam` file should then be updated in two places:
```

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (alias react)
+ (libraries lib reason-react melange-fetch)
+ (preprocess
+  (pps reactjs-jsx-ppx))
+ (module_systems es6))
+```
+
+### NPM packages
+
+A number of Melange compatible packages can be found on npm. Many older, but still useful, compatible BuckleScript libraries can be found on npm, e.g., `bs-json`. An npm package is installed like any normal npm package is: `npm install @glennsl/bs-json`, which will add the package to the `package.json` file at the root of our project.
```
anmonteiro, line 356:
```suggestion
A number of Melange compatible packages can be found on npm. Many older, but still useful, compatible BuckleScript libraries can be found on npm, e.g., `bs-json`. Run `npm install @glennsl/bs-json` to add the dependency locally and record it in the `package.json` file at the root of our project.
```

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+In order to get Dune to pick up and process the npm package so that it can be consumed in our application, the `dune` file at the root of our project must be updated with a `subdir` stanza:
```
anmonteiro, line 358:
```suggestion
Dune needs to be made aware of the newly installed package. The [`subdir`](https://dune.readthedocs.io/en/stable/dune-files.html#subdir) stanza can be handy in these cases:
```

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+```text
+(subdir
+ node_modules
+ (vendored_dirs @glennsl)
```
anmonteiro, line 363: remove vendored dirs. it won't work like this because we're not defining public libraries.

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (dirs @glennsl)
+ (subdir
```
anmonteiro, line 365: you can save a few lines here with `(subdir @glennsl/bs-json/src ... )`

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+   (subdir
+    src
+    (library
+     (name bs_json)
+     (wrapped false)
+     (modes melange))))))
+```
+
+If the `dune` file contains the line `(dirs :standard \ node_modules)`, it should be removed.
+Note the `(vendored_dirs @glennsl)` field in the stanza. This is used to silence warnings from vendored libraries, keeping the terminal output relevant to our application code.
```
anmonteiro, line 378: this is wrong in the example we give because we're not defining a public library in the vendored directory.

docs/package-management.md
```diff
@@ -267,3 +267,127 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+```text
+(library
+ (name data)
+ (libraries bs_json)
+ (modes melange))
+```
+
+Note that the library `bs-json` was renamed to `bs_json` in the `subdir` stanza and is referenced as `bs_json` in the `dune` file.
+
+If we install any more npm packages then a new `subdir` stanza will have to be added to the `dune` file in the root of our project. See [this dune file](https://github.com/jchavarri/pupilfirst/blob/b6fe6fa27814336b9e92f965449f3e010e148c4b/dune) for a larger example that uses multiple npm packages.
```
anmonteiro, line 391:
```suggestion
We can add new `subdir` stanzas for every package we'd like to consume this way. See [this dune file](https://github.com/jchavarri/pupilfirst/blob/b6fe6fa27814336b9e92f965449f3e010e148c4b/dune) for a larger example that uses multiple npm packages.
```

anmonteiro (commented):
Great start, @psb!

I added some comments.

Did you also run this example? it feels to me like the `vendored_dirs` part doesn't really work because there's no public library being defined, but I could be wrong. In any case, we should make sure the examples we're giving have been tested locally

docs/package-management.md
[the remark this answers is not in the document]
psb: I know nothing about publishing packages, so I can't really write anything about it.

docs/package-management.md
psb: I've changed the text to:

> Some libraries will only work after being processed by an accompanying PPX. For example, `reason-react` requires preprocessing with `reactjs-jsx-ppx`. These preprocessors may be installed together with the library as part of the same package, or they might be part of a different package, in which case they need to be installed separately.

psb:
> Did you also run this example? it feels to me like the `vendored_dirs` part doesn't really work because there's no public library being defined, but I could be wrong. In any case, we should make sure the examples we're giving have been tested locally

The example is from an app I'm running locally and it does work; however, it also works if I remove the `vendored_dirs`, which I did after seeing your comment.

I don't really understand when I would know if I should use `vendored_dirs` or not and whether a public library is/has been defined. Playing around with my dune file, bs-json and bs-fetch don't require `vendored_dirs` but bs-webapi does, otherwise I get lots of warnings in my terminal. Hopefully, I'll come to understand this better over time.

docs/package-management.md
psb: > the other advantage is that you only have to build them once, (i.e. if you vendor libraries you need to build them as part of your project)

I don't really know what this means. Would it be best to remove this line from this document and it to a document about publishing Melange packages?

docs/package-management.md
jchavarri: @anmonteiro we have been using `opam` all over, because that is how it is written in their doc page. https://opam.ocaml.org/

docs/package-management.md
jchavarri: 😕

docs/package-management.md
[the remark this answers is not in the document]
anmonteiro: It's OK to leave it off of this PR.

docs/package-management.md
anmonteiro: Probably best to remove it, good idea.

docs/package-management.md
anmonteiro: ah fair point. I wasn't aware.

docs/package-management.md
anmonteiro: sorry this is my mistake.

anmonteiro:
Seems like it was my mistake.

The `vendored_dirs` problem I mentioned only happens if the vendored directory has a `dune-project` file (effectively marking a different project, from which you can't consume private libraries).

It's probably simpler to omit it for now.

anmonteiro:
Sorry about the back and forth on `OPAM` vs `opam`. Let's use the lowercase version like @jchavarri [mentioned (comment)](https://github.com/melange-re/melange-re.github.io/pull/83#discussion_r1233082596).

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
 inside the bindings— is that it gives users of the bindings complete flexibility
 over the way these JavaScript packages are downloaded and bundled.
 
+## Finding and using Melange compatible packages
+
+### opam packages
+
+Melange packages are usually available on [opam](https://opam.ocaml.org). Package search is available in the opam CLI via `opam search <package-name>`, e.g. `opam search reason-react`. You can run `opam install <package-name>` to download, build and install opam packages in your switch. Remember that opam won't automatically add the dependency to `<your-project>.opam` file, so it must be added manually:
```
jchavarri, line 274: > Melange packages are usually available on opam.

This is a bit of a stretch 😬  as I think only 1 Melange package is available in the official opam repository today (`reason-react`). But that's the ultimate goal, and we have to be optimistic 🚀  that over time more packages will be published, so I would leave it as is.

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+<pre class="text-ocaml"><code class="language-text hljs plaintext">project_name/
+├── _opam
+├── src
+│   ├── dune
+│   ├── ReactComponent1.ml
+│   ├── ReactComponent2.ml
+│   └── lib
+│        ├── dune
+│        └── data.ml
```
jchavarri, line 294: ~Is such a complex file tree including a `lib/dune` file necessary to illustrate the example?~ Edit: oh, never mind, I see it's used below to show the `bs-json` use case.

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+(melange.emit
+ (target output)
+ (alias react)
+ (libraries lib reason-react)
+ (preprocess
+  (pps reactjs-jsx-ppx))
+ (module_systems es6))
+```
+
+Some libraries will only work after being processed by an accompanying PPX, e.g., `reason-react` requires preprocessing with `reactjs-jsx-ppx`. These preprocessors may be installed together with the library as part of the same package, or they might be part of a different package, in which case they need to be installed separately.
```
jchavarri, line 325: The `reactjs-jsx-ppx` might be renamed soon, we will have to remember to update the docs here.

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+ (dirs @glennsl)
+ (subdir
+  @glennsl/bs-json/src
+  (library
+   (name bs_json)
+   (wrapped false)
+   (modes melange)))
+```
+
+If the `dune` file contains the line `(dirs :standard \ node_modules)`, it should be removed.
```
jchavarri, line 372:
```suggestion
If the `dune` file contains the line `(dirs :standard \ node_modules)`, it should be removed, so that Dune can process the new Melange sources under the `node_modules` folder.
```

Not sure if we should also link to the ordered set language section in [the dune docs](https://dune.readthedocs.io/en/stable/reference/ordered-set-language.html).

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+In our project structure above we have the file <code class="text-ocaml">data.ml</code><code class="text-reasonml">data.re</code> under the folder `src/lib`. If we want use the `bs-json` library from within the <code class="text-ocaml">data.ml</code><code class="text-reasonml">data.re</code> file then we need to add the library name to the `dune` file in the same folder, i.e., `src/lib/dune`:
+
+```text
+(library
+ (name data)
+ (libraries bs_json)
+ (modes melange))
+```
+
+Note that the library `bs-json` was renamed to `bs_json` in the `subdir` stanza and is referenced as `bs_json` in the `dune` file.
```
jchavarri, line 383: Do you think we should add more context about why this is needed? e.g.

```suggestion
Note that the library `bs-json` was defined as `bs_json` in the `subdir` stanza and is referenced as `bs_json` in the `dune` file. This is necessary as Dune wrapped libraries [will only expose a single top-level module](https://dune.readthedocs.io/en/stable/explanation/ocaml-ecosystem.html#dune-is-opinionated) named after the library, so the library name has to be a valid module name. This is why library names with characters like `-` are not valid.
```

docs/package-management.md
```diff
@@ -267,3 +267,119 @@ The advantage of this approach —as opposed to vendoring the JavaScript package
...
+
+We can add new `subdir` stanzas for every package we'd like to consume this way. See [this dune file](https://github.com/psb/reason-react-hn-melange/blob/main/dune) for a larger example that uses multiple npm packages.
```
jchavarri, line 385: Better to link to the commit hash in case the file moves around or changes in the future:

```suggestion
We can add new `subdir` stanzas for every package we'd like to consume this way. See [this dune file](https://github.com/psb/reason-react-hn-melange/blob/9983b26aebd78b445d5f44d66bb6781eccedc787/dune) for a larger example that uses multiple npm packages.
```

jchavarri (approved):
This is looking great, thanks again for taking the time. I left a few minor suggestions, once @anmonteiro 👍 as well, it should be good to go.

anmonteiro:
Thanks

PR #99, opened 2023-08-17
Set up a nix shell for the playground project

anmonteiro:


anmonteiro:
better in https://github.com/melange-re/melange-re.github.io/pull/214

PR #111, opened 2023-09-01
howto: migrate rescript lib

### LINKED ISSUE #105: Adding a section for: "Migrating a BuckleScript library to Melange"
Recently helped some Discord user where was struggling with a compile error, and turns out he missed the reactjs-jsx-ppx, `(wrapped false)` and other sneaky pieces.

We made the process 90 times in Ahrefs and would be a nice tutorial to help others contribute to melange-community.

jchavarri:
Fixes #105.

[Preview](https://github.com/melange-re/melange-re.github.io/blob/8fecfa205a68f191bfea0514663965100d5ebb96/docs/how-to-guides.md).

### DIFF
--- a/docs/how-to-guides.md
+++ b/docs/how-to-guides.md
@@ -1,0 +1,383 @@
+# How-to guides 
+
+## Migrate a ReScript library to Melange
+
+It is possible to use existing [ReScript](https://rescript-lang.org/) (formerly
+BuckleScript) code with Melange, mostly as is. However, as both projects evolve
+in different directions over time, it will become more challenging to do so as
+time goes by, as some of the most recent features of ReScript might not be
+directly convertible to make them work with Melange.
+
+For this reason, the recommendation is to migrate libraries at a time where they
+were compatible with past versions of ReScript, for example v9 or at least v10.
+
+These are the steps to follow:
+
+- Add an `opam` file
+- Add a `dune-project` file
+- Replace the `bsconfig.json` file with one or multiple `dune` files
+- (Optional) Migrate from ReScript syntax to Reason or OCaml syntaxes
+- Make sure everything works: `dune build @all`
+- Final step: remove `bsconfig.json` and adapt `package.json`
+
+Let's go through them in detail:
+
+### Add an `opam` file
+
+To migrate your ReScript library to Melange, you will need some packages.
+Melange is designed to be used with [opam](opam.ocaml.org/), the package manager
+of OCaml, which is explained in [its own section](./package-management.md).
+
+To get started with the library migration, let's create an `opam` file in your
+library's root folder with the minimum set of packages to start working:
+
+```text
+opam-version: "2.0"
+synopsis: "My Melange library"
+description: "A library for Melange"
+maintainer: ["<your_name>"]
+authors: ["<your_name>"]
+license: "XXX"
+homepage: "https://github.com/your/project"
+bug-reports: "https://github.com/your/project/issues"
+depends: [
+  "ocaml"
+  "dune"
+  "melange"
+]
+dev-repo: "git+https://github.com/your/project.git"
+```
+
+If your library was using [Reason syntax](https://reasonml.github.io/en/) (`re`
+files), you will need to add `"reason"` to the list of dependencies. If the
+library was using ReScript syntax (`res` files), you will need to add
+`rescript-syntax` to the list of dependencies. You can read more about how to
+migrate from ReScript syntax in the section below.
+
+At this point, we can create a [local opam
+switch](https://opam.ocaml.org/blog/opam-local-switches/) to start working on
+our library:
+
+```bash
+opam switch create . 5.1.0~rc2 -y --deps-only
+```
+
+Once this step is done, we can call `dune` from the library folder, but first we
+need some configuration files.
+
+### Add a `dune-project` file
+
+Create a file named `dune-project` in the library root folder. This file will
+tell Dune a few things about our project configuration:
+
+```text
+(lang dune 3.8)
+
+(using melange 0.1)
+```
+
+### Replace the `bsconfig.json` file with one or multiple `dune` files
+
+Now, we need to add a `dune` file where we will tell Dune about our library. You
+can put this new file next to the library sources, it will look something like
+this:
+
+```text
+(library
+ (name lib)
+ (modes melange))
+```
+
+Let's see how the most common configurations in `bsconfig.json` (or
+`rescript.json`) map to `dune` files. You can find more information about these
+configurations in the [Rescript
+docs](https://rescript-lang.org/docs/manual/latest/build-configuration) and in
+the [Dune docs](https://dune.readthedocs.io/en/stable/dune-files.html#library).
+
+#### `name`, `namespace`
+
+These two configurations map to Dune `(wrapped <boolean>)` field in the
+`library` stanza. By default, all Dune libraries are wrapped, which means that a
+single module with the name of the library is exposed at the top level. So e.g.
+of your `bsconfig.json` had `"namespace": false`, you can add `(wrapped false)`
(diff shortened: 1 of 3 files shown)


docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+## Migrate a ReScript library to Melange
+
+It is possible to use existing [ReScript](https://rescript-lang.org/) (formerly
+BuckleScript) code with Melange, mostly as is. However, as both projects evolve
+in different directions over time, it will become more challenging to do so as
+time goes by, as some of the most recent features of ReScript might not be
+directly convertible to make them work with Melange.
+
+For this reason, the recommendation is to migrate libraries at a time where they
+were compatible with past versions of ReScript, for example v9 or at least v10.
```
anmonteiro, line 12:
```suggestion
were compatible with past versions of ReScript, for example v9 (v10 at most).
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+
+These are the steps to follow:
+
+- Add an `opam` file
```
anmonteiro, line 16: this is redundant if the dune-project files generates opam files

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+- Add a `dune-project` file
+- Replace the `bsconfig.json` file with one or multiple `dune` files
+- (Optional) Migrate from ReScript syntax to Reason or OCaml syntaxes
+- Make sure everything works: `dune build @all`
```
anmonteiro, line 20: `dune build @all` == `dune build`, so we can avoid potentially confusing people who don't know aliases yet

```suggestion
- Make sure everything works: `dune build`
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+ (modes melange))
+```
+
+Let's see how the most common configurations in `bsconfig.json` (or
+`rescript.json`) map to `dune` files. You can find more information about these
+configurations in the [Rescript
+docs](https://rescript-lang.org/docs/manual/latest/build-configuration) and in
+the [Dune docs](https://dune.readthedocs.io/en/stable/dune-files.html#library).
+
+#### `name`, `namespace`
```
anmonteiro, line 97: we should document that `"namespace": "foo-bar"` should translate to `(library (name fooBar))`

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+- `(dirs foo bar)`
+  ([docs](https://dune.readthedocs.io/en/stable/dune-files.html#dirs)): This
+  stanza tells Dune to only look into `foo` and `bar` subdirectories of the
+  current folder.
+
+So for example, if your library had this configuration in its `bsconfig.json`:
+
+```json
+{
+  "sources": ["src", "lib"]
```
anmonteiro, line 129: pick something other than `lib`, since bucklescript uses that for artifacts

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+opam 2.2. For example, to add `melange-jest` as dev dependency, you could add
+this in your library `opam` file:
+
+```text
+depends: [
+  "melange-jest" {with-dev-setup}
+]
+```
+
+Then, the library `melange-jest` would become available to your library tests.
```
anmonteiro, line 193: I suppose this is a bit of a weird example because we'd most likely want `{with-test}` in this case.

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+`dune-project` files can be added across a single codebase to work in a monorepo
+setup.
+
+#### `external-stdlib`
+
+There is no direct mapping of this functionality in Melange. If you are
+interested in it, or have a use case for it, please share with us on [issue
+melange-re/melange#620](https://github.com/melange-re/melange/issues/620).
+
+#### `js-post-build`
```
anmonteiro, line 212: maybe add an example of a rule that does this?

```clj
(rule 
  (deps (alias melange))
 ...
)
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+```bash
+rescript_syntax ./myFile.res -print ml | refmt --parse=ml --print re > myFile.re
+```
+
+Note that `refmt` is available in the `reason` package, so if your library
+modules are written using Reason syntax, remember to install it first using
+`opam install reason` before performing the conversion, and also adding it to
+your library `opam` file as well.
+
+### Make sure everything works: `dune build @all`
```
anmonteiro, line 308:
```suggestion
### Make sure everything works: `dune build`
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+
+Once you have performed the above steps, you can test that everything works by
+running
+
+```bash
+dune build @all
```
anmonteiro, line 314:
```suggestion
dune build
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+the type system in OCaml 4.12 was improved. You can read more about this in this
+[OCaml PR](https://github.com/ocaml/ocaml/pull/9783).
+
+**Fix**: either add `()` as final param of the function, or replace one labelled
+arg with a positional one.
+
+#### Warning 69 [unused-field] triggered from bindings types
+
+Sometimes, types for bindings will trigger `(warning 69 [unused-field]): record
+field sitekey is never read.` errors.
```
anmonteiro, line 332:
```suggestion
field foo is never read.` errors.
```

docs/how-to-guides.md
```diff
@@ -0,0 +1,383 @@
...
+**Fix**: Use `Stdlib`
+
+#### Runtime assets are missing
+
+In ReScript, building in source is very common. In Melange and Dune, the most
+common setup is having all artifacts inside the `_build` folder. If your library
+is using some asset such as:
+
+```ocaml
+external bkgImage : string = "default" [@@bs.module "./icons/overview.svg"]
```
anmonteiro, line 363:
```suggestion
external myImage : string = "default" [@@bs.module "./icons/overview.svg"]
```

anmonteiro (commented):
very nice guide, thanks!

docs/how-to-guides.md
jchavarri: True, but users might not have `dune` available on the global switch. I always run into this "chicken and egg" problem when using local switches, so I think it's simpler to create the `opam` file first, then the local switch, then proceed with Dune config. Wdyt?

docs/how-to-guides.md
jchavarri: Hm right great point. I will change this to `with-test` instead (which is what `bs-dev-deps` is mainly used for).

docs/how-to-guides.md
jchavarri: https://github.com/melange-re/melange-re.github.io/pull/111/commits/1f40d9cbf2f68fca8bd47770107d74845520041a

docs/how-to-guides.md
jchavarri: https://github.com/melange-re/melange-re.github.io/pull/111/commits/e60d446cff11f95d6061577f40ec31165018e949

docs/how-to-guides.md
jchavarri: https://github.com/melange-re/melange-re.github.io/pull/111/commits/274f1b660349ad9bfa95397ab01bcb984b3a31c0

docs/how-to-guides.md
jchavarri: https://github.com/melange-re/melange-re.github.io/pull/111/commits/c1f2057e2b9acfef1d35a59fc8a1487d54383c8a

PR #113, opened 2023-09-05
"Migrate from v1 to v2" guide

### LINKED ISSUE #87: Add a small guide "Migrating from v1 to v2"
- v2 only compatible with OCaml 5.1
- Adding `melange.ppx` is now in 99% cases a must, as it has to be added in multiple cases where it was not needed before
- Some warnings have been turned to alerts, so they might be visible even if using `vendored_dirs`
- `Js_foo` modules need to be replaced with `Js.Foo` (same for `Belt_*` modules) 
- `Belt_MapInt` is now `Belt.Map.Int`
- The payload of `deriving` now has different type, `{ jsConverter = newType }` becomes `jsConverter { newType }` (see below)
- `bs.deriving` is replaced with `deriving`
- All `@bs` attrs are replaced with `@mel`
- Except `[@bs]` which is replaced by `[@u]`
- `@bs.val` is deprecated (can be removed)
- Effect handlers will be turned off
- Adding `melange.dom` and `melange.node` to libraries is required when using those modules.
- tbc...

jchavarri:
Fixes #87.

### DIFF
--- a/docs/how-to-guides.md
+++ b/docs/how-to-guides.md
@@ -513,7 +513,7 @@ In Ocaml syntax:
 | Before  | After |
 |---------------|---------------|
 | `[@@bs.deriving { jsConverter =  newType  }]` | `[@@deriving  jsConverter {  newType }  ]` |
-| `[@@deriving { abstract = light }]` | `[@@deriving abstract { light }]` |
+| `[@@bs.deriving { abstract = light }]` | `[@@deriving abstract { light }]` |
 
 
 In Reason syntax:
@@ -521,7 +521,7 @@ In Reason syntax:
 | Before  | After |
 |---------------|---------------|
 | `[@bs.deriving {jsConverter: newType}]` | `[@deriving jsConverter({newType: newType})]`  |
-| `[@deriving {abstract: light}]` | `[@deriving abstract({light: light})]`  |
+| `[@bs.deriving {abstract: light}]` | `[@deriving abstract({light: light})]`  |
 
 #### `bs.*` attributes and extensions become `mel.*`
 
--- a/docs/how-to-guides.md
+++ b/docs/how-to-guides.md
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
 dependencies needed by Melange from the `package.json`, as they will be
 appearing in the `opam` file instead, as it was mentioned in the
 [`bs-dependencies` section](#bs-dependencies).
+
+## Migrate
+
+This section contains information about migrating from older versions of Melange
+to newer ones.
+
+### To v2 from v1
+
+Melange v2 is only compatible with OCaml 5.1. In order to upgrade, let's update
+the local opam switch first, to make sure the local repository gets the versions
+v2 of Melange and 5.1 of OCaml:
+
+```bash
+opam update
+```
+
+Now, update the version of the OCaml compiler in the local switch to 5.1:
+
+```bash
+opam install --update-invariant ocaml-base-compiler.5.1.0
+```
+
+Finally, we can upgrade all packages to get Melange v2 and the latest version of
+all libraries:
+
+```bash
+opam upgrade
+```
+
+To make sure you have the latest version of Melange, you can use the `opam list`
+subcommand:
+
+```bash
+opam list --installed melange
+# Packages matching: name-match(melange) & installed
+# Name  # Installed    # Synopsis
+melange 2.0.0          Toolchain to produce JS from Reason/OCaml
+```
+
+Before building, we have to update some parts of the configuration to make it
+work with v2.
+
+#### `melange.ppx` includes now most syntax transformations
+
+Most of the attributes used to write bindings are now handled by `melange.ppx`.
+If you get errors of the kind `Unused attribute`, then you probably need to add
+`melange.ppx` to your `library` or ` melange.emit` stanzas.
+
+```
+(library
+ ...
+ (preprocess
+  (pps melange.ppx)))
+```
+
+#### Warnings have been turned to alerts
+
+Some warnings were turned into alerts, so they might be visible even if using
+`vendored_dirs`. To silence these alerts, either fix the root cause or silence
+them using `melange.compile_flags`.
+
+#### Wrapped libraries
+
+Melange libraries like Belt and Js are now wrapped, so the access to modules
+inside them need to be adapted. Some examples:
+
+- `Js_foo` modules need to be replaced with `Js.Foo` (same for `Belt_*` modules) 
+- `Belt_MapInt` is now `Belt.Map.Int`
+
+#### Changes in `deriving`
+
+The `bs.deriving` attribute is replaced with `deriving`. Also, the payload taken
+by this attribute has been adapted to conform to ppxlib requirements.
+
+Here is a table showing how the payload has changed in both OCaml and Reason
+syntaxes:
+
+| Syntax | Before  | After |
+|---------------|---------------|---------------|
+| `ml`  | `[@@bs.deriving { jsConverter =  newType  }]` | `[@@deriving  jsConverter {  newType }  ]` |
+| `re`  | `[@bs.deriving {jsConverter: newType}]` | `[@deriving jsConverter({newType: newType})]`  |
+| `ml`  | `[@@deriving { abstract = light }]` | `[@@deriving abstract { light }]` |
+| `re`  | `[@deriving {abstract: light}]` | `[@deriving abstract({light: light})]`  |
+
+#### Changes in `bs` prefix for attributes and extension nodes
+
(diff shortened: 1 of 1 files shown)


docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+#### `Dom` and `Node` are in their own libraries
+
+The namespaces `Dom` and `Node` are now in the libraries `melange.dom` and
+`melange.node` respectively. These libraries are not included by default by
+Melange, and will have to be added to the `libraries` field to enable their
+usage.
+
+#### Effects handlers 
+
+???
```
jchavarri, line 543: @anmonteiro I am not sure what we want to mention here, if anything. Could you give me a hand please?

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+melange 2.0.0          Toolchain to produce JS from Reason/OCaml
+```
+
+Before building, we have to update some parts of the configuration to make it
+work with v2.
+
+#### `melange.ppx` includes now most syntax transformations
+
+Most of the attributes used to write bindings are now handled by `melange.ppx`.
+If you get errors of the kind `Unused attribute`, then you probably need to add
```
anmonteiro, line 478: this will also be true for "most weird errors you can't explain". it's probably just missing the ppx

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+`melange.ppx` to your `library` or ` melange.emit` stanzas.
+
+```
+(library
+ ...
+ (preprocess
+  (pps melange.ppx)))
+```
+
+#### Warnings have been turned to alerts
```
anmonteiro, line 488:
```suggestion
#### Warnings have been turned into alerts
```

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+Some warnings were turned into alerts, so they might be visible even if using
+`vendored_dirs`. To silence these alerts, either fix the root cause or silence
+them using `melange.compile_flags`.
```
anmonteiro, line 492: this is false. it doesn't go in `melange.compile_flags`. check https://github.com/melange-re/melange/blob/d6359b4607c50b8c31b93e8625202f9f6c51d7aa/test/blackbox-tests/ppx-alert.t#L12

e.g.:
```
(preprocess (pps melange.ppx -alert -deprecated))
```

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+#### Wrapped libraries
+
+Melange libraries like Belt and Js are now wrapped, so the access to modules
+inside them need to be adapted. Some examples:
+
+- `Js_foo` modules need to be replaced with `Js.Foo` (same for `Belt_*` modules)
```
anmonteiro, line 499: can you give a specific example instead of `Js_foo`?

anmonteiro: if we mention Belt in the line below I think we don't need the parentheses in this line

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+Melange libraries like Belt and Js are now wrapped, so the access to modules
+inside them need to be adapted. Some examples:
+
+- `Js_foo` modules need to be replaced with `Js.Foo` (same for `Belt_*` modules) 
+- `Belt_MapInt` is now `Belt.Map.Int`
+
+#### Changes in `deriving`
+
+The `bs.deriving` attribute is replaced with `deriving`. Also, the payload taken
```
anmonteiro, line 504: can we mention explicitly that `mel.deriving` won't be accepted?

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+|---------------|---------------|---------------|
+| `ml`  | `[@@bs.deriving { jsConverter =  newType  }]` | `[@@deriving  jsConverter {  newType }  ]` |
+| `re`  | `[@bs.deriving {jsConverter: newType}]` | `[@deriving jsConverter({newType: newType})]`  |
+| `ml`  | `[@@deriving { abstract = light }]` | `[@@deriving abstract { light }]` |
+| `re`  | `[@deriving {abstract: light}]` | `[@deriving abstract({light: light})]`  |
+
+#### Changes in `bs` prefix for attributes and extension nodes
+
+All attributes or extension nodes prefixed with `bs` are now prefixed with
+`mel`.
```
anmonteiro, line 520:
```suggestion
`mel` instead.
```

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+For example `@bs.as` becomes `@mel.as`, and `%bs.raw` becomes `%mel.raw`. The
+only exception is the `@bs` attribute, which becomes `@u`, for uncurried
```
anmonteiro, line 523: I think `@u` deserves its own section

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+| `re`  | `[@bs.deriving {jsConverter: newType}]` | `[@deriving jsConverter({newType: newType})]`  |
+| `ml`  | `[@@deriving { abstract = light }]` | `[@@deriving abstract { light }]` |
+| `re`  | `[@deriving {abstract: light}]` | `[@deriving abstract({light: light})]`  |
+
+#### Changes in `bs` prefix for attributes and extension nodes
+
+All attributes or extension nodes prefixed with `bs` are now prefixed with
+`mel`.
+
+For example `@bs.as` becomes `@mel.as`, and `%bs.raw` becomes `%mel.raw`. The
```
anmonteiro, line 522: we should mention that `@bs.*` is still accepted until v3, but `%bs.*` isn't

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+`external` definitions. See more information in the ["Using global functions or
+values"](./communicate-with-javascript.md#using-global-functions-or-values)
+section.
+
+#### `Dom` and `Node` are in their own libraries
+
+The namespaces `Dom` and `Node` are now in the libraries `melange.dom` and
+`melange.node` respectively. These libraries are not included by default by
+Melange, and will have to be added to the `libraries` field to enable their
+usage.
```
anmonteiro, line 539:
```suggestion
Melange, and will need to be added to the `libraries` field explicitly.
```

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+#### Effects handlers
```
anmonteiro, line 541:
```suggestion
#### Effect handlers 
```

docs/how-to-guides.md
anmonteiro: I suggest we say something along these lines:

> Although Melange v2 requires OCaml 5.1, we don't yet have a good solution for compiling effect handlers to JavaScript. Until we do, they are disabled at the compiler level, and their APIs are not accessible to users.

jchavarri:
@anmonteiro thanks for the review, i think i tackled all the suggestions in the last couple of commits.

docs/how-to-guides.md
```diff
@@ -430,3 +430,114 @@ If everything went well, you can remove the `bsconfig.json` file, and remove any
...
+
+#### Changes in `deriving`
+
+The `bs.deriving` attribute is replaced with `deriving`. Also, the payload taken
+by this attribute has been adapted to conform to ppxlib requirements.
+
+Here is a table showing how the payload has changed in both OCaml and Reason
+syntaxes:
+
+| Syntax | Before  | After |
```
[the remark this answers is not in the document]
jchavarri: We would need to split the `Before` column as well. I ended up splitting in two tables:

https://github.com/melange-re/melange-re.github.io/blob/fddb06db3e07db43a1e2f52ebcd14deb4972178d/docs/how-to-guides.md?plain=1#L511-L524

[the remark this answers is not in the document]
jchavarri: Noticed some inconsistent naming in the "Before" column, fixed in https://github.com/melange-re/melange-re.github.io/pull/113/commits/19dfd3d5bdd7cf51232f537dd5e0381cb008212a.

PR #120, opened 2023-10-03
Bindings cookbook

jchavarri:
@rusty-key suggested to include a "cookbook" section in the bindings page, so that users can grep for common JS patterns and map them to bindings.

This PR adds a cookbook at the end of "Communicate with JavaScript" page, based on https://github.com/yawaramin/bucklescript-bindings-cookbook.

### DIFF
--- a/docs/communicate-with-javascript.md
+++ b/docs/communicate-with-javascript.md
@@ -3166,11 +3166,11 @@ let default = 10;
 That way, Melange will set the value on the `default` export so it can be
 consumed as default import on the JavaScript side.
 
-### Bindings cookbook
+## Bindings cookbook
 
-#### Globals
+### Globals
 
-##### `window`: global variable
+#### `window`: global variable
 
 ```ocaml
 external window : Dom.window = "window"
@@ -3182,7 +3182,7 @@ external window: Dom.window = "window";
 See the [Using global functions or values](#using-global-functions-or-values)
 section for more information.
 
-##### `window?`: does global variable exist
+#### `window?`: does global variable exist
 
 ```ocaml
 match [%mel.external window] with
@@ -3199,7 +3199,7 @@ switch ([%mel.external window]) {
 See the [Detect global variables](#detect-global-variables) section for more
 information.
 
-##### `Math.PI`: variable in global module
+#### `Math.PI`: variable in global module
 
 ```ocaml
 external pi : float = "PI" [@@mel.scope "Math"]
@@ -3212,7 +3212,7 @@ See the [Binding to properties inside a module or
 global](#binding-to-properties-inside-a-module-or-global) section for more
 information.
 
-##### `console.log`: function in global module
+#### `console.log`: function in global module
 
 ```ocaml
 external log : 'a -> unit = "log" [@@mel.scope "console"]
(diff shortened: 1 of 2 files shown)


docs/communicate-with-javascript.md
```diff
@@ -3154,3 +3165,622 @@ let default = 10;
...
+module Foo = {
+  module Bar = {
+    [@mel.module "foo"] [@mel.scope "bar"] external baz: unit => unit = "baz";
+  };
+};
+
+let () = Foo.Bar.baz();
+```
+
+It is not necessary to nest the binding inside Reason modules, but mirroring the
```
schinns, line 3297:
```suggestion
It is not necessary to nest the binding inside Reason/OCaml modules, but mirroring the
```

This wording could cause confusion to the reader that there is a differentiation between Reason and OCaml modules? We could also just say 'OCaml' modules?

docs/communicate-with-javascript.md
```diff
@@ -3154,3 +3165,622 @@ let default = 10;
...
+```reasonml
+external range: (~start: int, ~stop: int, ~step: int=?, unit) => array(int) =
+  "range";
+
+let nums = range(~start=1, ~stop=10, ());
+```
+
+When an OCaml function takes an optional parameter, it needs a positional
+parameter at the end of the parameter list to help the compiler understand when
+function application is finished and the function can actually execute. This
```
schinns, line 3371:
```suggestion
function application is finished and when the function can actually execute. This
```

docs/communicate-with-javascript.md
```diff
@@ -3154,3 +3165,622 @@ let default = 10;
...
+might seem cumbersome, but it is necessary to have out-of-the-box curried
+parameters, named parameters and optional parameters available in the language.
```
schinns, line 3373:
```suggestion
might seem cumbersome, but it is necessary in order to have out-of-the-box curried
parameters, named parameters, and optional parameters available in the language.
```

docs/communicate-with-javascript.md
```diff
@@ -3154,3 +3165,622 @@ let default = 10;
...
+external foo: Foo.t = "foo";
+
+let _result =
+  switch (Foo.bar(foo)) {
+  | Some(_) => 1
+  | None => 0
+  };
+```
+
+If you know some value may be `undefined` (but not `null`, see next section),
```
schinns, line 3744:
```suggestion
If you know some value may be `undefined` (but not `null`), see next section,
```

schinns (approved):
This is guide is amazing! Thanks @jchavarri. I made some minor suggestions.

docs/communicate-with-javascript.md
jchavarri: I believe this change modifies the meaning. The original meaning was that if the value may be `null`, then the reader should see next section. After this change that meaning is lost.

jchavarri (commented):
thanks for the review @schinns! I had a question about one of the suggestions.

docs/communicate-with-javascript.md
jchavarri: Good point, yeah let's use OCaml modules.

docs/communicate-with-javascript.md
maxkorp: Perhaps
```
If you know some value may be `undefined` (but not `null`, see the next section for more on `null`), and if you know...
```

feihong (approved):
lgtm

PR #122, opened 2023-10-04
Playground/type hints

### LINKED ISSUE #94: [Playground] Implement “type on-hover”
- Bundle stdlib / runtime .cmt files
- Generate .cmt from OCaml / ReasonML compilation

jchavarri:
Fixes #94.

### DIFF
--- a/playground/src/app.jsx
+++ b/playground/src/app.jsx
@@ -511,9 +511,8 @@ function App() {
 
   React.useEffect(() => {
     let hoverProvider = undefined;
-    // or make sure that it exists by other ways
     if (monaco) {
-      hoverProvider = monaco.languages.registerHoverProvider(languageMap.Reason, {
+      hoverProvider = monaco.languages.registerHoverProvider(language, {
         provideHover: function (model, position) {
           const { lineNumber, column } = position;
           if (!compilation?.typeHints) {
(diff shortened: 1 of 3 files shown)


playground/src/app.jsx
```diff
@@ -420,15 +420,36 @@ const compile = (language, code) => {
     }
     if (problems) {
       return {
+        typeHints: [],
         problems: problems,
       }
     } else {
       return {
+        typeHints: compilation.type_hints.sort((a, b) => {
```
davesnx, line 428: I believe sort mutates the array and returns the reference to the same array. You might want to be immutable here with `.slice().sort()`

davesnx (approved):

playground/src/app.jsx
```diff
@@ -488,6 +509,67 @@ function App() {
...
+                result.end.col + 1
+              );
+              let hint = result.hint;
+              if (language == languageMap.Reason) {
+                const prefix = "type t = ";
+                try {
+                  hint =
+                    ocaml
+                    /* add prefix so it is valid code */
+                    .printRE(ocaml.parseML(prefix + hint))
```
davesnx, line 546: is there any case where this can be problematic?

playground/src/app.jsx
jchavarri: yeah... i was aware but because we get a new hints array every time we recompile i was lazy and left the mutating version.

playground/src/app.jsx
jchavarri: where it can raise an exn? who knows... 🤷  just playing it safe.

davesnx: I mean the prefix `type = t` being added to the hint. Do recursive types might break here? or other cases?

jchavarri: Ah, good point! After your comment, I tested more thoroughly and found out this approach would break for type definitions, where we already have a valid AST and we can parse directly. I fixed it in https://github.com/melange-re/melange-re.github.io/pull/122/commits/fa07c2962c2baf6824ecef33e669768b3846d318.

PR #125, opened 2023-10-19
Update switch version

davesnx:


jchavarri:
Thanks! All appearances of OCaml versions are being updated in #124, including this one.

PR #165, opened 2024-02-12
getting-started: add create-melange-app

jchavarri:


docs/getting-started.md
```diff
@@ -1,8 +1,29 @@
 # Getting started
 
-Get up and running with Melange in no time!
+Get up and running with Melange in no time! Below you will find the two main
+ways to get started with Melange:
+- (Recommended) With
```
davesnx, line 5: I wouldn't call it recommended but rather automatic or out-of-the-box and if we recommend it happy to add it thought

davesnx (approved):

docs/getting-started.md
jchavarri: Updated in https://github.com/melange-re/melange-re.github.io/pull/165/commits/a7bdce3f14768add788367eaebd6456882934f2c.

PR #169, opened 2024-02-27
getting-started: add link to book

jchavarri:


docs/getting-started.md
```diff
@@ -1,7 +1,14 @@
 # Getting started
 
-Get up and running with Melange in no time! Below you will find the two main
-ways to get started with Melange:
+If you would like to learn OCaml and Melange from scratch, we recommend you to
+read ["Melange for React Devs"](https://react-book.melange.re/). This book will
+give you an overview of the OCaml language, as well as showcase some of the
+mechanisms that Melange offers to interact with JavaScript code. You’ll build a
```
feihong, line 6: Chapters on binding aren't ready yet. But I guess it'll be obvious when they click through and see that it's not in the table of topics on the intro page?

feihong (approved):

docs/getting-started.md
jchavarri: There are already quite a few mentions of interop features like `mel.module`, `mel.raw`, `Js.t` and runtime representation of records, so while an official "Bindings" chapter is not there yet, folks can already explore some of this functionality inside the book :)

PR #196, opened 2024-10-11
Add 'users of melange widget' to front page

feihong:


docs/.vitepress/theme/Users.vue
```diff
@@ -0,0 +1,70 @@
+<script>
+export const users = [
+  {
+    name: "Ahrefs",
+    image: "ahrefs.svg",
```
davesnx, line 5: Can we use the newest ahrefs logo from https://ahrefs.com/logo

docs/.vitepress/theme/index.js
```diff
@@ -1,10 +1,11 @@
 // You can directly import Vue files in the theme entry
 // VitePress is pre-configured with @vitejs/plugin-vue.
 import Layout from './Layout.vue'
```
davesnx, line 3: Just curious, but why do we use vue for this?

docs/.vitepress/theme/Users.vue
feihong: Updated in https://github.com/melange-re/melange-re.github.io/pull/196/commits/e51bf5ef1fa4ce7951ee044d9090c3bf4ba865de

docs/.vitepress/theme/index.js
feihong: Afaik, we have to use Vue components in VitePress if we want to use custom components. For example, @jchavarri implemented syntax switching by making a Vue component: https://github.com/melange-re/melange-re.github.io/blob/master/docs/.vitepress/theme/Switch.vue

jchavarri: Yes. There's no way around that, at least that I know of. Unless we move away from Vitepress to something React based.

docs/.vitepress/theme/Users.vue
```diff
@@ -0,0 +1,80 @@
...
+  },
+  {
+    name: "Qwick",
+    image: "qwick.svg",
+    infoLink: "https://qwick.com"
+  },
+  {
+    name: "Atvero",
+    image: "atvero.svg",
+    infoLink: "https://atvero.com",
```
jchavarri, line 16: This site doesn't open.

feihong: I guess it should be updated to https://www.cmap.io/ instead. Will contact the guy from Reason discord who works at CMap (which I assume acquired Atvero).

marcusroberts: Hello, just to confirm we did migrate the product to Melange, and the link you have is the correct one.  Thanks for updating!

davesnx:
Is it good to merge @feihong?

feihong:
@davesnx If you don't mind that there are only 5 companies right now, it can be merged

PR #203, opened 2024-10-30
Update reason snippets after 3.13

jchavarri:


docs/communicate-with-javascript.md
```diff
@@ -855,14 +855,13 @@ let add = [%mel.raw {|
 let () = Js.log (add 1 2)
 ```
 ```reasonml
-let add = [%mel.raw
-  {|
+let add =
+  {%mel.raw |
```
jchavarri, line 859: @anmonteiro This change is a bit strange, is it expected? If that's the way to write this now, we should probably adapt the references across the docs cc @feihong

feihong: Wow, this is huge change. I see no mention of it in https://github.com/reasonml/reason/blob/master/CHANGES.md.

@jchavarri This PR, when merged, will update Playground to Reason 3.13, then? Then I guess we should also change this line from 3.12 to 3.13:

https://github.com/melange-re/melange-re.github.io/blob/master/playground/src/app.jsx#L148

feihong (approved):
lgtm

docs/communicate-with-javascript.md
feihong: Is `{%mel.raw ||}` supported in OCaml syntax?

PR #211, opened 2024-12-06
Fix code snippet compile errors

feihong:
Also fix:
- [x] Formatting of reason snippets (because 3.14 formats records a bit differently)
- [x] Properly convert snippets that contain type signatures 

`process_md.ml` has been changed so that when it detects a snippet that begins with "val ", it assumes that it contains type signatures and wraps it with `module type _FAKE_  = sig ` and ` end`. After it converts to reason code, the extra lines are removed.

### DIFF
--- a/docs/working-with-js-objects-and-values.md
+++ b/docs/working-with-js-objects-and-values.md
@@ -202,7 +202,7 @@ manual](https://v2.ocaml.org/manual/objectexamples.html).
 We have already explored one approach for creating JavaScript object literals by
 using [`Js.t` values and the `mel.obj` extension](#using-js-t-objects).
 
-Melange additionally offers the `mel.obj` attribute, which can be used in
+Melange additionally offers the `@mel.obj` attribute, which can be used in
 combination with external functions to create JavaScript objects. When these
 functions are called, they generate objects with fields corresponding to the
 labeled arguments of the function.
@@ -215,47 +215,44 @@ optional keys are emitted at runtime.
 For example, assuming we need to bind to a JavaScript object like this:
 
 ```js
-var homeRoute = {
-  type: "GET",
-  path: "/",
-  action: () => console.log("Home"),
-  // options: ...
+let place = {
+  name: "Boring"
+  type: "city",
+  greeting: () => console.log("Howdy"),
+  // attractions: ...
 };
 ```
 
-The first three fields are required and the `options` field is optional. You can
-declare a binding function like:
+The first three fields are required and the `attractions` field is optional. You
+can declare a binding function like this:
 
 ```ocaml
-external route :
+external makePlace :
+  name:string ->
   _type:string ->
-  path:string ->
-  action:(string list -> unit) ->
-  ?options:'a ->
+  greeting:(unit -> unit) ->
+  ?attractions:string array ->
   unit ->
-  'b = ""
-  [@@mel.obj]
+  _ = ""
+[@@mel.obj]
 ```
 ```reasonml
 [@mel.obj]
-external route:
+external makePlace:
   (
+    ~name: string,
     ~_type: string,
-    ~path: string,
-    ~action: list(string) => unit,
-    ~options: 'a=?,
+    ~greeting: unit => unit,
+    ~attractions: array(string)=?,
     unit
   ) =>
-  'b;
+  _;
 ```
 
-Note that the empty string at the end of the function is used to make it
-syntactically valid. The value of this string is ignored by the compiler.
-
-Since there is an optional argument `options`, an additional unlabeled argument
-of type `unit` is included after it. It allows to omit the optional argument on
-function application. More information about labeled optional arguments can be
-found in the [OCaml
+Since there is an optional argument `attractions`, an additional unlabeled
+argument of type `unit` is included after it. It allows to omit the optional
+argument on function application. More information about labeled optional
+arguments can be found in the [OCaml
 manual](https://v2.ocaml.org/manual/lablexamples.html#s:optional-arguments).
 
 The return type of the function should be left unspecified using the wildcard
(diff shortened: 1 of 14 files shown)


davesnx (approved):

scripts/process_md.ml
```diff
@@ -7,14 +7,33 @@ let print_with f structureAndComments =
...
+  in
+  let lines = String.split_on_char '\n' s in
+  let len = List.length lines in
+  lines
+  |> List.filteri (fun i _ -> i <> 0 && i <> len - 2)
+  |> List.map dedent |> String.concat "\n"
+
 let to_re str =
+  let is_type_signature, str =
+    (* assume that any snippet starting with "val" is a partial type signature *)
```
jchavarri, line 24: Nice.

scripts/working-with-js-objects-and-values.t
```diff
@@ -71,16 +68,19 @@ file. To update the tests, run `dune build @extract-code-blocks`.
...
-  File "input.ml", line 2, characters 8-11:
-  2 | let _ = map [||] [||] add
-              ^^^
-  Error: Unbound value map
-  Hint: Did you mean max?
+  File "input.ml", line 5, characters 22-25:
+  5 | let _ = map [||] [||] add
+                            ^^^
+  Error: This expression has type int -> int -> int
+         but an expression was expected of type ('a -> 'b -> 'c [@u])
```
jchavarri, line 83: Is this expected?

jchavarri (approved):
Very nice. Thanks so much!

scripts/working-with-js-objects-and-values.t
feihong: Yes, it's expected. This example is supposed to not compile because `map` expects an uncurried function as its third argument, but `add` is curried.

It's sometimes hard to figure out which snippet is the one that's supposed to break because they aren't extracted in the order that they appear in the markdown file.

feihong: I added comments in the prelude of snippets that aren't expected to compile in https://github.com/melange-re/melange-re.github.io/pull/211/commits/f79420b4be9d8e32731235068d9bffa03bdf84e4

docs/working-with-js-objects-and-values.md
```diff
@@ -217,9 +231,9 @@ external route :
   _type:string ->
   path:string ->
   action:(string list -> unit) ->
-  ?options:< .. > ->
+  ?options:'a ->
```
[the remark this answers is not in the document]
feihong: I rewrote this example based on the feedback I got in our conversation: https://github.com/melange-re/melange-re.github.io/pull/211/commits/f13706e5404ad1dc11f3a2fcb16aa8b230e0e971

scripts/advanced-js-interop.t
```diff
@@ -182,17 +159,19 @@ file. To update the tests, run `dune build @extract-code-blocks`.
...
   $ dune build @melange
-  File "input.ml", line 1, characters 17-23:
-  1 | val actionToJs : action -> string
-                       ^^^^^^
-  Error: Unbound type constructor action
-  Hint: Did you mean option?
+  File "input.ml", line 4, characters 0-33:
+  4 | val actionToJs : action -> string
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Error: Value declarations are only allowed in signatures
```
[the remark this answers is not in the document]
feihong: Didn't fix it but added a comment to indicate that the snippet is a type signature in https://github.com/melange-re/melange-re.github.io/pull/211/commits/f79420b4be9d8e32731235068d9bffa03bdf84e4

docs/working-with-js-objects-and-values.md
```diff
@@ -188,7 +202,7 @@ manual](https://v2.ocaml.org/manual/objectexamples.html).
 We have already explored one approach for creating JavaScript object literals by
 using [`Js.t` values and the `mel.obj` extension](#using-js-t-objects).
 
-Melange additionally offers the `mel.obj` attribute, which can be used in
+Melange additionally offers the `@mel.obj` attribute, which can be used in
```
jchavarri, line 205: The problem with adding `@` here is that OCaml syntax uses two `@@` and Reason one `@`. Before it was syntax agnostic, but now it's not. Maybe we can use the "inline syntax switching" elements to work around this.

docs/working-with-js-objects-and-values.md
```diff
@@ -252,32 +263,78 @@ In the route function, the `_type` argument starts with an underscore. When
 binding to JavaScript objects with fields that are reserved keywords in OCaml,
 Melange allows the use of an underscore prefix for the labeled arguments. The
 resulting JavaScript object will have the underscore removed from the field
-names. This is only required for the `mel.obj` attribute, while for other cases,
-the `mel.as` attribute can be used to rename fields.
+names. This is only required for the `@mel.obj` attribute, while for other
+cases, the `@mel.as` attribute can be used to rename fields.
```
jchavarri, line 267: Same comments as above for inclusion of `@` and OCaml syntax.

feihong: Fixed in another PR: https://github.com/melange-re/melange-re.github.io/pull/212/commits/ba9bf1d433289f0823397bc4db0c128f02e8d6c3

docs/working-with-js-objects-and-values.md
feihong: Makes sense, fixed in another PR: https://github.com/melange-re/melange-re.github.io/pull/212/commits/ba9bf1d433289f0823397bc4db0c128f02e8d6c3

I was used to referring to annotations with `@` prefixed to the name because the book only uses Reason syntax, so there's no ambiguity.

PR #219, opened 2025-03-08
add melange 5 blogpost

anmonteiro:


blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
+---
+title: Announcing Melange 5
+date: 2025-03-05
```
anmonteiro, line 3: change to the desired publish date

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+});
+```
+
+### Dynamically `import()`ing JavaScript from OCaml
+
+One of Melange's primary operating principles is the ability to support
+seamless interop with JavaScript constructs. As such, we implemented `import()`
+in a way that also allows importing JS modules dynamically: you can call
+`Js.import` on JavaScript values declared with `external`. The abstractions
+compose nicely here to produce the expected result. Check out this example of
```
jchavarri, line 77:
```suggestion
compose nicely to produce the expected result. Check out this example of
```

Otherwise sounds like they don't compose nicely somewhere else 😄

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+1. **build system-compatible**: dynamic `import()`s in Melange work with Dune
+   out of the box: as usual, you must specify your `(library ..)` dependencies
+   in the `dune` file. At compile time, Melange will be aware of the
+   dynamically imported module locations to emit the arguments to
+   `import("/path/to/module.js)` automatically.
+2. **type-safe**: your OCaml code is still aware of its dependencies at
+   compile-time, but Melange will skip emitting static JavaScript `import ..`
+   declarations if the values are exclusively used through `Js.import(..)`.
+
+### Dynamically `import()`ing OCaml code
```
jchavarri, line 44:
```suggestion
### Dynamically importing OCaml code
```

Current version is a bit hard to read for me (same for other subtitle below). But maybe i'm getting old.

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+
+The example below makes it clear: we import the entire `Stdlib.Int` module,
+specify its type signature, and observe that no static `import`s appear in the
```
jchavarri, line 47: Should we clarify (here or somewhere else) that dynamic imports are an "unsafe" boundary, similar to bindings? Meaning, the developer is responsible of defining the right module type being imported, and the compiler has no way of checking if this definition is correct.

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+
+Starting from this release, we're shipping NPM packages with the precompiled
+Melange runtime. This feature, requested by a few users in
+[melange#620](https://github.com/melange-re/melange/issues/620) allows to use
+Melange without compiling its own runtime and stdlib (essentially, in
+combination with `(emit_stdlib false)` in `(melange.emit ..)`).
+
+This can be useful in monorepos that compile multiple Melange applications but,
+perhaps most importantly, it enables Melange libraries and packages to be
+published without the weight of the full runtime / stdlib.
```
jchavarri, line 239:
```suggestion
perhaps most importantly, it enables Melange libraries and packages to also
be published in NPM without the weight of the full runtime / stdlib.
```

(if that is that is being implicitly explained?)

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Melange runtime NPM packages
+
+Starting from this release, we're shipping NPM packages with the precompiled
+Melange runtime. This feature, requested by a few users in
+[melange#620](https://github.com/melange-re/melange/issues/620) allows to use
+Melange without compiling its own runtime and stdlib (essentially, in
+combination with `(emit_stdlib false)` in `(melange.emit ..)`).
+
+This can be useful in monorepos that compile multiple Melange applications but,
```
jchavarri, line 237: This use case is still unclear to me. Assuming a single opam switch, why would one want to add npm overhead (and potential mismatch of melange versions definitions between opam and npm)?

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+inception](https://anmonteiro.com/2021/03/on-ocaml-and-the-js-platform/), one
+of its goals has been to keep it up to date with the latest OCaml releases.
+[This
+release](https://github.com/melange-re/melange/releases/download/5.0.1-53/melange-5.0.1-53.tbz)
+brings Melange up to speed with OCaml 5.3, including upgrades to the `Stdlib`
+library as well. We're also releasing Melange 5 for OCaml
+[4.14](https://github.com/melange-re/melange/releases/download/5.0.1-414/melange-5.0.1-414.tbz),
+[5.1](https://github.com/melange-re/melange/releases/download/5.0.1-51/melange-5.0.1-51.tbz)
+and
+[5.2](https://github.com/melange-re/melange/releases/download/5.0.1-52/melange-5.0.1-52.tbz).
```
jchavarri, line 227: ❤️  This is just 🤯

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+`push` method, which produces the following JavaScript:
+
+```js
+[].push(3);
+```
+
+Besides being more versatile, having an explicit marker with `@mel.this` is
+also more visually intuitive: when scanning Melange code containing `external`
+bindings, it becomes easier to spot which is the "this" argument. This feature
+is 100% backwards compatible with `@mel.send`: in the absence of `@mel.this`,
```
jchavarri, line 209: The 100% read strange, either a feature like this is backwards compatible or it's not 😇

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Better editor support for Melange `external`s
+
+Melange bindings to JavaScript, specified through `external` declarations, used
+to propagate internal information in the [native
+payload](https://ocaml.org/manual/5.3/intfc.html#external-declaration). In
+practice, hovering over one of these in your editor could end up looking a bit
+weird:
+
+![](https://private-user-images.githubusercontent.com/661909/386930735-af1c5c90-2ba7-4813-b4d8-80a4f9a74aae.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDEzOTE4MjIsIm5iZiI6MTc0MTM5MTUyMiwicGF0aCI6Ii82NjE5MDkvMzg2OTMwNzM1LWFmMWM1YzkwLTJiYTctNDgxMy1iNGQ4LTgwYTRmOWE3NGFhZS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwMzA3JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDMwN1QyMzUyMDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0xYTVmNDA3Nzk4MjhlNGEzNjE0MmVkMGVlOTM3NDgzMzM4MzUwYWVlNjJjNmMxMmY1NjU2Mjg4MjM2NGYyYTI0JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.T6wIwTPOmHyWpcCBcJoon1d42nQjMGzqch7P2BbRI3Q)
```
jchavarri, line 249: I am not able to see these images in the rendered version. I think we'll have to create some `blog/posts/images` folder and use relative links for this to work?

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+The example below makes it clear: we import the entire `Stdlib.Int` module,
+specify its type signature, and observe that no static `import`s appear in the
+resulting JavaScript:
+
+```js
+// dynamic_import_int.re
+
+module type int = (module type of Int);
+
+let _: Js. Promise.t(unit) = {
```
jchavarri, line 55:
```suggestion
let _: Js.Promise.t(unit) = {
```

jchavarri (approved):
Super well written, this is a great upgrade 🌟 Thanks so much for your hard work on this and the blog post!

I think the only blocker are the images not being visible, otherwise it's just small nits.

blog/posts/announcing-melange-5.md
anmonteiro: that's actually wrong :) it's fully type-safe, and the compiler would complain if `module Int` wasn't of `module type of int`!

blog/posts/announcing-melange-5.md
```diff
@@ -0,0 +1,285 @@
...
+
+```js
+// dynamic_import_int.js
+
+import("melange/int.js").then(function (Int) {
+  return Promise.resolve((console.log(Int.max), undefined));
+});
+```
+
+### Dynamically `import()`ing JavaScript from OCaml
```
anmonteiro, line 71:
```suggestion
### Dynamically importing JavaScript from OCaml
```

PR #222, opened 2025-04-30
Add a bindings cheatsheet and reason-react binding examples

johnhaley81:
I thought having a big simple page for quickly looking up binding examples would be useful. This is a completely new page with no header/sidebar.

Also added some reason-react binding examples.

cc @davesnx @rusty-key

### DIFF
--- a/docs/public/melange-cheatsheet.html
+++ b/docs/public/melange-cheatsheet.html
@@ -1,0 +1,676 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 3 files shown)


docs/public/melange-cheatsheet.html
```diff
@@ -0,0 +1,676 @@
...
+      display: block;
+    }
+
+    .syntax__reasonml .reasonml {
+      display: block;
+    }
+  </style>
+</head>
+
+<body>
```
davesnx, line 110: is there any chance to use the regular header here? or treat this page as a .md file, etc?

Re-creating everything might cause future updates to be more painful

davesnx (commented):
Any chance we can keep the snippets in sync? I would love to not mess with HTML and just update a list and have both updated

